Issue #4, opened 2018-04-14
Widget: Text block


mum4k:
Implement a simple text block that with styling and wrapping options.

Similar to gizak/termui: https://github.com/gizak/termui/raw/master/_example/par.png

mum4k:
Also support rolling text.

Similar to: https://github.com/yaronn/blessed-contrib/raw/master/docs/images/log.gif

PR #35, opened 2018-05-14
[4/5] Adding an object that tracks the scrolling state.

mum4k:
Works on #4


widgets/text/scroll.go
```diff
@@ -0,0 +1,147 @@
...
+// line scroll out of the view, so the content rolling is paused.
+func rollingPaused(st *scrollTracker, lines, height int) rollState {
+	st.first = st.doScroll(lines, height)
+	if lastLineVisible(st.first, lines, height) {
+		return rollToEnd
+	}
+	return rollingPaused
+}
+
+// lastLineVisible returns true if the last line is visible given drawing that
```
chipmunkie: Could we re-write this comment to clarify what the last line means. Is it the last line in the buffer?

widgets/text/scroll.go
```diff
@@ -0,0 +1,147 @@
...
+// This is not thread safe.
+type scrollTracker struct {
+	// scroll stores user requests to scroll up (negative) or down (positive).
+	// E.g. -1 means up by one line and 2 means down by two lines.
+	scroll int
+
+	// scrollPage stores user requests to scroll up (negative) or down
+	// (positive) by a page of content. E.g. -1 means up by one page and 2
+	// means down by two pages.
+	scrollPage int
```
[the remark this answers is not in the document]
mum4k: I have expanded the comment for the object itself which now hopefully covers this.

widgets/text/scroll.go
mum4k: Comment rewritten. Please let me know if this doesn't help.

PR #36, opened 2018-05-14
[5/5] Implementing the text widget.

mum4k:


widgets/text/text.go
```diff
@@ -0,0 +1,281 @@
...
+// Reset resets the widget back to empty content.
+func (t *Text) Reset() {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+
+	t.buff.Reset()
+	t.givenWOpts = newGivenWOpts()
+	t.scroll = newScrollTracker(t.opts)
+	t.lastWidth = 0
+	t.newText = true
```
chipmunkie: Should this be true if there is no content and has been reset (reset kind of gives the impression on new text was added yet).

widgets/text/text.go
```diff
@@ -0,0 +1,281 @@
...
+		if cur.Y == height-1 && height >= minLinesForMarkers && height < lines-fromLine {
+			if height >= minLinesForMarkers {
+				if err := cvs.SetCell(cur, '⇩'); err != nil {
+					return err
+				}
+			}
+			break
+		}
+
+		if r == '\n' {
```
chipmunkie: Do we actually need this statement?

widgets/text/text.go
mum4k: Good point, renamed newText to contentChanged.

widgets/text/text.go
mum4k: Yes, otherwise we print out the newline rune below which adds an empty line of text.

Note - I am not happy with this function as it is in this pull request. I have refactored it into something hopefully easier to understand in the next one.

Issue #83, opened 2019-01-22
Cleanup code issues on go report card


mum4k:
https://goreportcard.com/report/github.com/mum4k/termdash

Note: not all of the gocyclo issues should be cleaned up, some functions need to have those switch statements with many cases (like selecting a character). Similarly we don't really want to shorten tests. So the gocyclo category needs judgement.

The other categories are straightforward.

infinytum:
I've created a pull-request #84 which fixes all except the go-cyclo errors

PR #84, opened 2019-01-22
Partial Fix for #83 (Excluding Gocyclo)

infinytum:
I fixed the errors except for gocyclo. Its an A+ grade now

### DIFF
--- a/draw/line_style.go
+++ b/draw/line_style.go
@@ -39,19 +39,6 @@ var lineStyleChars = map[LineStyle]map[linePart]rune{
 		vAndRight:         '├',
 		vAndH:             '┼',
 	},
-	LineStyleDouble: {
-		hLine:             '═',
-		vLine:             '║',
-		topLeftCorner:     '╔',
-		topRightCorner:    '╗',
-		bottomLeftCorner:  '╚',
-		bottomRightCorner: '╝',
-		hAndUp:            '╩',
-		hAndDown:          '╦',
-		vAndLeft:          '╣',
-		vAndRight:         '╠',
-		vAndH:             '╬',
-	},
 }
 
 // init verifies that all line parts are half-width runes (occupy only one
@@ -88,15 +75,13 @@ func (ls LineStyle) String() string {
 
 // lineStyleNames maps LineStyle values to human readable names.
 var lineStyleNames = map[LineStyle]string{
-	LineStyleLight:  "LineStyleLight",
-	LineStyleDouble: "LineStyleDouble",
+	LineStyleLight: "LineStyleLight",
 }
 
 // Supported line styles.
 const (
 	LineStyleNone LineStyle = iota
 	LineStyleLight
-	LineStyleDouble
 )
 
 // linePart identifies individual line parts.
--- a/eventqueue/eventqueue_test.go
+++ b/eventqueue/eventqueue_test.go
@@ -107,7 +107,7 @@ func TestPullBlocksUntilAvailable(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Millisecond)
 	defer cancel()
 
-	if _, err := q.Pull(ctx); err != nil {
+	if _, err := q.Pull(ctx); err == nil {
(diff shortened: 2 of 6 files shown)


eventqueue/eventqueue_test.go
```diff
@@ -107,13 +107,13 @@ func TestPullBlocksUntilAvailable(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Millisecond)
 	defer cancel()
 
-	got, err := q.Pull(ctx)
+	_, err := q.Pull(ctx)
```
mum4k: Go usually prefers limited scopes for variables. This is especially true about the err variable which gets used a lot. Limiting its scope prevents programming errors where we might forget to check it.

Now that we don't use the return value, we can limit the scope of err to the if statement:
if _, err := q.Pull(ctx); err != nil {
    t.Fatal("Pull => expected timeout error, got nil")
}

widgets/linechart/axes/label.go
```diff
@@ -89,7 +89,7 @@ func rowLabel(scale *YScale, y int, labelWidth int) (*Label, error) {
 	}
 
 	ar := rowLabelArea(y, labelWidth)
-	pos, err := align.Text(ar, v.Text(), align.HorizontalRight, align.VerticalMiddle)
+	pos, _ := align.Text(ar, v.Text(), align.HorizontalRight, align.VerticalMiddle)
```
mum4k: Thanks for fixing this, this was actually a bug. We however shouldn't silently discard the error variable as that can create a hard to debug scenario. I.e. the align.Text() function fails and reports an error, we ignore it and silently report success. The failure of align.Text() will most likely result in some unexpected behavior that will be hard to find.

If we correctly process and report the error, it will lead us to this very line when something goes wrong.

We can do something like this, adding additional context to the reported error:

pos, err := align.Text(ar, v.Text(), align.HorizontalRight, align.VerticalMiddle)
if err != nil {
    return nil, fmt.Errorf("unable to align the label value: %v", err)
}

eventqueue/eventqueue_test.go
infinytum: Alright, I fixed that. Please note that I adjusted your code  to match err == nil, as the original if statement did exactly that, and the tests would fail otherwise.

Thanks for the advice on in-if-statement checking!

widgets/linechart/axes/label.go
infinytum: Hehe, I actually fixed that once, then I decided to revert it again... It is fixed now too. Glad we found a bug here!

eventqueue/eventqueue_test.go
mum4k: Of course, this was a case of "tired eyes" on my end. Thank you.

draw/line_style.go
```diff
@@ -39,6 +39,19 @@ var lineStyleChars = map[LineStyle]map[linePart]rune{
 		vAndRight:         '├',
 		vAndH:             '┼',
 	},
+	LineStyleDouble: {
```
mum4k: This looks great, thanks a lot!

Would you mind sending this in a separate pull request? A couple of reasons really:
1) this being a new feature - I would prefer to merge it into the devel branch instead of the master and follow a release cycle.
2) for the posterity - the pull request history or rollbacks become much cleaner when each pull request is focused on only one bug / feature / etc.
3) we could iterate on this change a bit more without blocking your original PR regarding the lint issues. This new feature (new line style) could benefit from some test coverage. An advice I got a long time ago was to accompany each code change with at least minimal corresponding change in tests to "capture" the new feature in a working state.

I have no more comments on this PR without the new line style, so once we remove the line style from this scope - this is ready to go.

For (3) we could expand the test cases inside the border_test:

https://github.com/mum4k/termdash/blob/master/draw/border_test.go

infinytum: I've reverted the merge with the line-style. 

I'll check out that test file 👍

mum4k: Thank you.

PR #102, opened 2019-02-07
Release 0.6.0

mum4k:
### Added

- The SegmentDisplay widget.
- A CHANGELOG.
- New line styles for borders.

### Changed

- Better recordings of the individual demos.

### Fixed

- The LineChart now has an option to change the behavior of the Y axis from
  zero anchored to adaptive.
- Lint errors reported on the Go report card.
- Widgets now correctly handle a race when new user data are supplied between
  calls to their Options() and Draw() methods.


mum4k:
On CLA: This is a pull between two internal branches, the other contributors were OK with contributing to the devel branch which was done from their own pull requests.

PR #133, opened 2019-02-18
The LineChart widget now supports zooming the content

### LINKED ISSUE #55: Support zooming of the LineChart.
Rotation of the mouse wheel or key presses could adjust the scale.

mum4k:
Fixes #55


keithknott26:
@mum4k,
This looks great!!!
I confirmed it works on both a Trackpad and Mouse w/ scroll wheel. Clicking anywhere in the linegraph box zooms 100% even if the Linegraph doesn't have "focus" indicated by the yellow border.

One very minor thing I noticed was that when you don't apply the XUnscaled option and continually fill the Series with records, scrolling will eventually start, for me it happens at about 1450 records in the Series. I'm sure you put this in as a feature but wanted to mention it just in case you didnt. 

The other very minor thing I noticed was under that same scenario (XUnscaled option enabled) and after scrolling starts, you can see the left most label changing to be the right most label.  (1:11 vs 00:35)

I've included a screenshot of the behavior, I can attach a screen recording if you like. The interesting thing is it doesn't happen each time the graph is scrolled to the left, it happens every 1 out of 100 or so notches and ONLY happens when the graph is completely zoomed out. Almost like when the function that handles the scrolling might have a bug.

image

Edit: I'll get a screen recording together and post it here shortly.

mum4k:
Thanks for the detailed testing @keithknott26, very helpful.

Note that the **focus** only applies to keyboard events, i.e. the infrastructure directs the keyboard events to the focused widget only. Mouse events don't have the same restriction because we know the coordinates of the mouse click.

Thinking about the **zoom too 100%** on a single click made me realize how misleading this can be. The intention is to make the user select a range (click and drag, then release). The fact that it zooms after a single click (single row highlighted) can be confusing. I have filed #135 and will fix it next.

It wasn't my intention to allow the graph to scroll when XAxisUnscaled isn't set. I have tried to reproduce this, but even after pushing 3000+ values, the X axis is still firmly based at zero. I waited until 5000+ and it still didn't happen.

image

However I believe this is related to zooming, see #136. I was also able to reproduce the behavior of the first label being misplaced and believe it is related to #136.

mum4k:
With both #135 and #136 resolved, this should all behave better.

Both fixes pushed to the devel branch.

keithknott26:
#135 and #136 confirmed through testing. I'm not able to find any other issues -  I'll research a bit more and see if that bug where the line starts scrolling is on my end.

mum4k:
Great, fyi I found one more bug (#140) which shows itself when we try to zoom into a linechart that rolls (and keep it zoomed until the zoomed portion rolls out of the view).

The fix is already on its way to devel.

PR #185, opened 2019-04-14
Allow empty values on line chart

### LINKED ISSUE #184: Allow "no value" on lineChart
Hi!

Would be nice to allow no values on the linechart, at this moment the linechart only supports float64 values (that defaults to `0` if no value) but sometimes you don't have values to render on the graph. A  `0` and a `nil` are two different things and could make the graph *break* by the Y axis getting very big (from 0 to N) in very high value graphs without very differences.

An example to show the graph when we don't have values:

We have this graph on Prometheus that shows us that from 5:27 to 5:29 doesn't have values (not rendered and the Y axis doesn't show `0`):

Same as above but rendered using termdash:

The problem with the termdash rendering is that the "no values" are rendered with the default float64 as `0` and the Y axis shows `0` making the scale of the graph bigger and by this harder to get insights from it.

The first approach to solve this that comes to my mind is by having `*float64` as value unit or similar (custom type with pointers on no values) in another method like `SeriesWithNils`, `SeriesWithNoValues` so it doesn't break the current API.

slok:
This PR Fixes #184.

It will allow having _values_ on the line chart that will not be displayed if they are missing. To do so it will use the `math.NaN` float64 value.

If this is combined with adaptive Y-axis option, the graphs that have empty gaps and high values with small variance will be more accurate regarding the Y axis (because they will not start at 0).

In this example, we can see that we don't have metrics around 10:03-10:05 making the graph render values as 0 and making Y axis start at 0:

Instead of representing the missing metrics as 0, with this change and using `math.NaN`, the Y axis has a better range around the min and max values making the graph more accurate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The grid.Builder API now allows users to specify options for intermediate
   containers, i.e. containers that don't have widgets, but represent rows and
   columns.
+- Line chart widget now allows `math.NaN` values to represent "no value" (values
+  that will not be rendered) in the values slice.
 
 #### Breaking API changes
 
@@ -85,10 +87,10 @@ identifiers shouldn't be used externally.
 - The draw.LineStyle enum was refactored into its own package
   linestyle.LineStyle. Users will have to replace:
 
-  -  draw.LineStyleNone -> linestyle.None
-  -  draw.LineStyleLight -> linestyle.Light
-  -  draw.LineStyleDouble -> linestyle.Double
-  -  draw.LineStyleRound -> linestyle.Round
+  - draw.LineStyleNone -> linestyle.None
+  - draw.LineStyleLight -> linestyle.Light
+  - draw.LineStyleDouble -> linestyle.Double
+  - draw.LineStyleRound -> linestyle.Round
 
 ## [0.7.0] - 24-Feb-2019
 
@@ -121,7 +123,6 @@ identifiers shouldn't be used externally.
 
 - The Text widget now has a Write option that atomically replaces the entire
   text content.
-
 
 #### Improvements to the infrastructure
 
@@ -251,7 +252,7 @@ identifiers shouldn't be used externally.
 - The Gauge widget.
 - The Text widget.
 
-[Unreleased]: https://github.com/mum4k/termdash/compare/v0.8.0...devel
+[unreleased]: https://github.com/mum4k/termdash/compare/v0.8.0...devel
 [0.8.0]: https://github.com/mum4k/termdash/compare/v0.7.2...v0.8.0
 [0.7.2]: https://github.com/mum4k/termdash/compare/v0.7.1...v0.7.2
 [0.7.1]: https://github.com/mum4k/termdash/compare/v0.7.0...v0.7.1
--- a/internal/numbers/numbers.go
+++ b/internal/numbers/numbers.go
@@ -109,14 +109,18 @@ func Round(x float64) float64 {
 
 // MinMax returns the smallest and the largest value among the provided values.
 // Returns (0, 0) if there are no values.
+// Ignores NaN values.
 func MinMax(values []float64) (min, max float64) {
 	if len(values) == 0 {
 		return 0, 0
 	}
 	min = math.MaxFloat64
 	max = -1 * math.MaxFloat64
-
 	for _, v := range values {
+		if math.IsNaN(v) {
+			continue
+		}
+
 		if v < min {
 			min = v
 		}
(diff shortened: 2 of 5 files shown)


internal/numbers/numbers.go
```diff
@@ -109,14 +109,18 @@ func Round(x float64) float64 {
...
+// Ignores NaN values.
 func MinMax(values []float64) (min, max float64) {
 	if len(values) == 0 {
 		return 0, 0
 	}
 	min = math.MaxFloat64
 	max = -1 * math.MaxFloat64
-
 	for _, v := range values {
+		if math.IsNaN(v) {
```
mum4k: I think we might have a corner case now - if all the values are NaN, this function will return min==math.MaxFloat64, max==-1 * math.MaxFloat64.

Wondering what the right thing to do is, but considering that the min and max cannot be determined from NaN values, should we return min==NaN, max==NaN?

We probably should also document the corner case in the doc comment and add a corresponding test case.

widgets/linechart/linechart_test.go
```diff
@@ -960,6 +960,53 @@ func TestLineChartDraws(t *testing.T) {
 				return ft
 			},
 		},
+		{
+			desc:   "more values than capacity, X rescales with NaN values ignored",
```
mum4k: There are a couple more interesting cases we might want to test:

- all the values are NaN.
- the very first visible value is NaN.
- the very last visible value is NaN.

I would appreciate if you would consider adding the test cases. However I know this is a lot of work, so if you want you can just confirm that the corner cases work and I will be happy to add the missing test coverage after we merge your PR in.

If you do choose to add the tests, we could consider adding them cases with less values in the series (just the smallest amount we need for the test case). That should make them easier to write.

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The grid.Builder API now allows users to specify options for intermediate
   containers, i.e. containers that don't have widgets, but represent rows and
   columns.
+- Line chart widget now allows `math.NaN` values to represent "no value" (values
```
mum4k: One more request in relation to the CLA - can you please explicitly confirm on this PR that I have your permission to merge this into the master branch in the next release?

internal/numbers/numbers.go
slok: You are right, maybe we could return some default values for min and max in this case, for example (-1,1), or (0,1).

widgets/linechart/linechart_test.go
slok: You are right, no problem! I'll add them.

CHANGELOG.md
slok: @mum4k has permission to merge this PR in master branch in the next release.

internal/numbers/numbers.go
mum4k: Thinking about this more - this might not be the right place to do that. This function is fairly generic "numerical" function and has more users than just the LineChart. I do agree with you that the behavior of LineChart should be to set some default min and max if all values are NaN.

How about we keep this function generic and return min==NaN, max==NaN when all values are NaN to indicate to all possible callers that the min and max cannot be determined? And then the LineChart internally can interpret the min==NaN, max==NaN as min==0, max==0. Note the LineChart should handle both min and max equal to zero.

slok:
Ready for a new check!

I've added 2 things as you noted:

- `numbers.MinMax` now returns `NaN min and max` when all values are NaN.
- LineChart defaults to 0 when the min or max is NaN.

widgets/linechart/linechart.go
```diff
@@ -519,3 +531,22 @@ func (lc *LineChart) maxXValue() int {
 	}
 	return maxLen - 1
 }
+
+const (
```
mum4k: (optional) Since these consts are only used within the minMax function below, we could as well define them inside the function. Encapsulating them in might give a better explanation of what they are for.

Or we could as well not use the consts and just go with literal zeroes, up to you.

internal/numbers/numbers_test.go
```diff
@@ -215,12 +215,29 @@ func TestMinMax(t *testing.T) {
...
+			wantMax: math.NaN(),
+		},
 	}
 
 	for _, tc := range tests {
 		t.Run(tc.desc, func(t *testing.T) {
 			gotMin, gotMax := MinMax(tc.values)
-			if gotMin != tc.wantMin || gotMax != tc.wantMax {
+			// Different assertion for NaN cases.
+			if (math.IsNaN(tc.wantMin) && !math.IsNaN(gotMin)) ||
```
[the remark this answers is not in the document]
slok: Much better. I wasn't sure if you wanted to use something different than standard conditionals to check floats.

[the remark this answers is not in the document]
mum4k: I think pretty is fine even for simple types assuming it works, we use it in majority of tests in this project already.

slok:
Ready for a new review!

Sorry for all the time is requiring this PR, thank you.

mum4k:
Looks good, thank you for all the work on this PR!

Issue #95, opened 2019-01-27
LineChart should limit the use of decimal digits on the Y axes 


mum4k:
The following screenshot demonstrates the issue:

<img width="644" alt="screen shot 2019-01-26 at 20 30 23" src="https://user-images.githubusercontent.com/5315344/51794869-62a9de00-21a9-11e9-9240-1ff25385b22e.png">

All the values are integers in the range of 1600-1800, yet the LineChart shows floating point values. This is because when the LineChart calculates the value of each cell row (called the step) it ends up being a floating point number.

Step of the Y axis is calculated here:
https://github.com/mum4k/termdash/blob/1e1dc15c47adf691d0f5a698a19d0662f2b5ab14/widgets/linechart/axes/scale.go#L65

The labels are determined here:
https://github.com/mum4k/termdash/blob/1e1dc15c47adf691d0f5a698a19d0662f2b5ab14/widgets/linechart/axes/label.go#L85

We could add a conditional and if the step between two adjacent rows is larger than X (say 10 or 20?) the generated label will rounded to the nearest integer.

Issue #104, opened 2019-02-09
Support floating / pop-up windows


mum4k:
Allow creation of a new window at a screen location. This window could internally use another instance of container to reuse the existing widget drawing code.

This could be use to interact with the user, letting them press buttons, show additional data or deliver messages.

PR #188, opened 2019-04-25
Y-axis label formatter for LineChart

### LINKED ISSUE #187: Support Y-axis custom labels
Hi!

At this moment, X-axis supports a slice of labels to customize what is rendered on the X-axis.

I haven't seen this for the Y-axis. If this is not implemented, would be nice to have the same feature for the Y-axis. The motivation behind this is to have the ability to do something similar to this:

A _simple_ (without having too much knowledge about the internals of the library) possible solution could be to allow setting a "transforming" function that accepts the `y` axis value/number as an argument and returns the string to render on the Axis. This way we could convert the Y value to any unit and display the units also.

What are your thoughts on this? would be this possible?

Thank you!

slok:
This PR fixes #187 

Adds the ability to set custom labels on the Y-axis using a value formatter that accepts a numeric value (float64) and returns the corresponding label to this value (string).

The PR is in WIP state. I'll update the PR accordingly.

Steps:
- [x] Set the Linechart Y-axis option with the `ValueFormatter`.
- [x] Use the formatter on the `Value` if is present.
- [x] Create formatter for most common time units.
- [x] Create formatter to round values (#95).

This approach seems to work fine, I put an image of the project that I'm developing (a Grafana for terminal that I'll make it public soon) using this feature.


slok:
Hi @mum4k!

I've added the first step of defining the API (I have a first implementation but I would like your feedback on the API option before pushing those changes to the PR).

mum4k:
Hi @slok, this all looks good, I don't have any comments at this time. Looks like we are on the same page.

Also I had no idea we could do draft PRs on GitHub, thank you for teaching me something new and for making the review easy by splitting it to small pieces.

slok:
Hi @mum4k, I just pushed the code that implements the usage of the formatter option.

I've been struggling a bit (did some iterations and I was not happy with the results haha) on the design of how to pass the formatter around, the approach that I selected is the least invasive I could think of (without knowing the design decisions you have taken of the linechart internals).

I would have preferred to remove `Value.Rounded`, `Value.ZeroDecimals` and `Value.NonZeroDecimals` struct attributes, and use a default formatter that takes the `Value.Value` and calculates this rounded float, but I've seen that `Value.Rounded` is not only used as a way of text representation but also used on internal logic of length, positions... so I didn't want to alter how this works.

widgets/linechart/linechart_test.go
```diff
@@ -1636,6 +1637,48 @@ func TestLineChartDraws(t *testing.T) {
 				testdraw.MustBrailleLine(bc, image.Point{10, 2}, image.Point{11, 0})
 				testbraille.MustCopyTo(bc, c)
 
+				testcanvas.MustApply(c, ft)
+				return ft
+			},
+		},
+		{
+			desc:   "custom Y-axis labels using a value formatter",
```
mum4k: When it comes to test coverage, we probably have a few more interesting cases. And one of them will be a bit more complicated because well unicode. We probably want to test:

1. A case where the formatter returns an empty string for the label.

2. A case where the formatter returns a string that is so long that it won't even fit on the canvas. Which the widget should handle by giving up in a graceful way. Hopefully it would just somehow hit this code which simply draws a character we are using to indicate "there isn't enough space":
https://github.com/mum4k/termdash/blob/a9515f27215c8c7f3c0df731c4e86103cb9cdb9d/widgets/linechart/linechart.go#L303-L305

3. Is the harder one. Previously we generated labels internally, so we had full control over them. We could guarantee they will only contain printable ASCII characters. Now the user could include any non-printable (control or space) character or any unicode character.

3a. **Non-printable characters** - open to opinions here. My gut tells me that we should document they aren't allowed and return an error from Draw() if a label would contain one. We can use the following function that Termdash already uses in a few locations to validate the label:
https://github.com/mum4k/termdash/blob/a9515f27215c8c7f3c0df731c4e86103cb9cdb9d/internal/wrap/wrap.go#L61-L81

3b. **Unicode** - the trick with unicode is that some unicode runes take more than one terminal cell to draw. Some examples are here:
https://github.com/mum4k/termdash/blob/a9515f27215c8c7f3c0df731c4e86103cb9cdb9d/internal/runewidth/runewidth_test.go#L46-L48

We should support these as it is a goal of Termdash to be Unicode ready. When these characters (or graphemes) are in a string, the strings length no longer equals its width on the screen. At the very least this code isn't going to work anymore:
https://github.com/mum4k/termdash/blob/a9515f27215c8c7f3c0df731c4e86103cb9cdb9d/widgets/linechart/internal/axes/axes.go#L125-L134

We will have to update it to use runewidth.StringWidth instead of len() when determining the width of a label. The addition of the test case might uncover other issues.

widgets/linechart/internal/axes/value.go
```diff
@@ -72,14 +84,24 @@ func (v *Value) Text() string {
...
-		return fmt.Sprintf("%.0f", v.Rounded)
+
+	if v.Formatter != nil {
+		return v.Formatter(v.Value)
+	}
+
+	return v.defaultFormatter(v.Rounded)
+}
+
+func (v *Value) defaultFormatter(value float64) string {
```
mum4k: (nit) Looks like this doesn't need to be a method anymore (it doesn't use the receiver), can it be a plain function?

```go
func defaultFormatter(value float64) string { ... }
```

widgets/linechart/internal/axes/value.go
```diff
@@ -49,12 +54,19 @@ func (v *Value) String() string {
 // NewValue returns a new instance representing the provided value, rounding
 // the value up to the specified number of non-zero decimal places.
 func NewValue(v float64, nonZeroDecimals int) *Value {
+	return NewFormattedValue(v, nonZeroDecimals, nil)
+}
+
+// NewFormattedValue returns a new instance representing the provided value,
+// using a value formatter.
+func NewFormattedValue(v float64, nonZeroDecimals int, formatter valueFormatter) *Value {
```
mum4k: (optional) Instead of creating a new constructor, we could add optional variadic arguments to NewValue which is what most of the Termdash repo does. Them being variadic means that none of the existing call sites would break.

Here is an example:
https://github.com/mum4k/termdash/blob/a9515f27215c8c7f3c0df731c4e86103cb9cdb9d/internal/draw/rectangle.go#L27-L65
What do you think?

mum4k: The privacy seems to be a bit mixed up here. I am guessing we had to define a separate type here to avoid circular dependencies, which is fine. However this is a public (exported from package) function which now accepts a private type that won't get documented in the exported GoDoc.

We might be changing this to variadic options in which case we can also choose to not define any type and just have an option accepting an anonymous function. This package is internal after all.

```go
func ValueFormatter(vf func(float64) string) ValueOption { ... }
```

Alternatively if you choose to keep the current approach, we should probably export and document the new type.

widgets/linechart/internal/axes/value.go
```diff
@@ -36,6 +38,9 @@ type Value struct {
 	// NonZeroDecimals indicates the rounding precision used, it is provided on
 	// a call to newValue.
 	NonZeroDecimals int
+	// Formatter will format value to a string representation of the value,
+	// if Formatter is not present it will fallback to default format.
+	Formatter valueFormatter
```
mum4k: Does this field need to be exported on the struct or can it be private? Will any of the call sites need to access it?

widgets/linechart/internal/axes/scale_test.go
```diff
@@ -22,8 +22,8 @@ import (
 )
 
 // mustNewYScale returns a new YScale or panics.
-func mustNewYScale(min, max float64, graphHeight, nonZeroDecimals int, mode YScaleMode) *YScale {
-	s, err := NewYScale(min, max, graphHeight, nonZeroDecimals, mode)
+func mustNewYScale(min, max float64, graphHeight, nonZeroDecimals int, mode YScaleMode, formatter valueFormatter) *YScale {
```
mum4k: This might still be work in progress, is the plan to use this and add one test that would set the valueFormatter? Or are we getting good test coverage via the other tests (label and axes)?

slok:
Wow! that was an awesome review, many thanks!

I'll check in detail, answer and solve in a few days. I don't have much time these coming days, sorry.

Best!

mum4k:
Thanks @slok, no worries feel free to get back to this whenever it suits. There are no time commitments for pet projects.

mum4k:
Hey @slok,

me and @keithknott26 were wondering how you got the legend under the LineChart in your screenshot up in this PR. Is this just a Text widget under the LineChart widget or are you planning to implement support for a legend?

Keith opened #191 so assuming you're not working on the legend, one of us (me or Keith) could work on supporting it in the LineChart directly.

Please let us know.

slok:
Hi @mum4k!

Yes, is composed of a lineachart and a text. On the app I'm creating a widget is an element that can be composed of multiple widgets/elements.

This way I can customize a legend that can be at any side and scrolls also.

Some examples:

mum4k:
Thanks @slok, that's a great usage of the Text widget. I will continue discussion around the legend in #191 in order to not derail this thread.

widgets/linechart/internal/axes/value.go
slok: Just to understand and learn, what's the benefit (or drawbacks) in these cases, converting a method to a function? I mean, for example, when I have these cases I maintain as an internal method to give the semantic meaning that this function is used internally by the struct and anything else.

widgets/linechart/internal/axes/value.go
slok: I thought that `Value` and `NewValue` was only used internally in the `axes` package and I tried following the same style of public factories although it was only used only on the same package.

It seemed strange to me but when I contribute to other people/group repos I try following the repos styles and decision whenever I can (IMHO uniformity is better in most cases) although sometimes I miss data for not being familiar with the repository as happened on this one, sorry.

widgets/linechart/internal/axes/value.go
mum4k: Happy to explain my position more and please feel free to ask about any comments if in doubt. It is indeed good to mark elements as internal for the reasons you explained. However a plain function can also be internal (unexported, start with lowercase letter), so this shouldn't affect the choice of function Vs. method.

Making function into a method extends the surface of its API by giving it an additional input or output; the receiver (v *Value). A reader of the code now assumes that the method either needs to read the object's state, or it might be modifying the object state as part of what it does. By making something that doesn't need to be a method a plain function, we are making the reader's life easier by being more explicit.

The argument I often make is about using the lowest primitive that does the job. One would not use a variable where a constant is enough, a function where a variable is enough or a method/object where a plain function is enough.

Of course there are languages that like to use objects as a way to package common elements together (like Java). Go's style tends to prefer clean object design as described above.

widgets/linechart/internal/axes/value.go
mum4k: No worries at all, honestly both ways of specifying the options are fine. The only concern is about not mixing public and private types in one declaration. Indeed NewValue is used from other packages as well (looks like zoom uses it). Otherwise the right choice would have been to make the functions private as well.

Thank you very much for taking the extra mile and trying to match local style, it indeed shows in your code and makes the review much easier.

widgets/linechart/internal/axes/value.go
slok: Understood, in those cases to let the user know that the struct will not be used I usually do this change `func (s SomeStruct) someMethod()` or ` func (s *SomeStruct) someMethod()` to `func (SomeStruct) someMethod()`. Although setting the plain func is more visual. Anyway, more than happy to set `defaultFormatter` as a plain function.

Thanks for the explanation!

widgets/linechart/internal/axes/scale_test.go
slok: We are already using it:
https://github.com/mum4k/termdash/pull/188/files#diff-82004fbfa0d341d2a0ce7179c86d5f13R200

widgets/linechart/linechart_test.go
slok: After adding the different test cases with the current code, these are the results:

- 1: Handles ok because it returns an error on the Draw, I think that for now, this could be correct although we could have an option in the future to disable labels on the different axis and this should not error.
- 2: Doesn't seem to reach that code:

```
               got:
                              │⢸
                              │⢸
                              │⢸
            51.680000000000000│⢸
                              │⡇
                              │⡇
                              │⡇
                             0│⡇
                              └─
                               0
              want:
                 │           ⢀⠔⠁
                 │          ⡠⠊
                 │        ⢀⠎
            61.7s│      ⢀⠔⠁
                 │     ⡠⠊
                 │   ⢀⠎
                 │ ⢀⠔⠁
                ∅│⡠⠊
                 └──────────────
                  0            1
```
I need to fix that case.

- 3.a: Without any change it returns a draw error, so we are ok also on this case.
- 3.b: It handles ok without any change, Do you think that this covers this case or I should use other unicode characters?

```
               got:                                                   
                 │           ⢀⠔⠁                                        
                 │          ⡠⠊                                          
                 │        ⢀⠎                                            
             ab世│      ⢀⠔⠁                                             
                 │     ⡠⠊                                               
                 │   ⢀⠎                                                
                 │ ⢀⠔⠁                                                 
              abc│⡠⠊                                          
                 └──────────────                              
                  0            1                                     
              want:                                              
                 │           ⢀⠔⠁                              
                 │          ⡠⠊                                
                 │        ⢀⠎                                  
             ab世│      ⢀⠔⠁                                          
                 │     ⡠⠊                                     
                 │   ⢀⠎                                                 
                 │ ⢀⠔⠁                                                  
              abc│⡠⠊                                                    
                 └──────────────                                        
                  0            1 
```
Although it seems to work without any change should I change `len` to use `runewidth.StringWidth`, I don't have enough experience with the lib to know if the Unicode chars are being checked in another code path along the way?

mum4k: Thanks a lot @slok for investigating this in detail.

(1) and (2) seem ok, I agree with your conclusions for (1) and I think the trimming behavior is most likely the best answer we can have for (2).

(3a) Looking at the code, I did not find any explicit validation other than checking for '\n' in this call stack:
https://github.com/mum4k/termdash/blob/e6af456c61cd50c650ae80857a74763d1ab4e158/widgets/linechart/internal/axes/label.go#L119-L122
https://github.com/mum4k/termdash/blob/e6af456c61cd50c650ae80857a74763d1ab4e158/internal/alignfor/align.go#L93-L96

Can you try with a formatter that returns one of these?
```go
// Space characters.
return "\ta"
// Control characters
fmt.Sprintf("%ca", 0x007f)
```

Basically characters that this test mentions ad invalid:
https://github.com/mum4k/termdash/blob/master/internal/wrap/wrap_test.go

Maybe when testing this, the chosen character triggered some other error? The other question we should ask is if the error returned from Draw was descriptive enough to make it clear what went wrong.

(3b) Looks like we will need to use runewidth.StringWidth and maybe also some other changes. Looking at the outputs above - it looks like we do have the bug, the unicode character caused an overflow into the next cell which shows itself as a loss of one space. A bit subtle, the unicode characters are tricky. Could you try with a string that will only have the wide characters? That should make the bug much more obvious as we will "lose" multiple cells.
```go
return "世世世世世世"
```

widgets/linechart/internal/axes/scale_test.go
mum4k: Thank you, I have somehow missed that.

slok:
Hey @mum4k, Ready for a new check!

Thanks for all your feedback and time, your tips, corner cases and pointing to the correct code paths have been a *HUGE*  help.

widgets/linechart/linechart_test.go
```diff
@@ -1679,6 +1679,133 @@ func TestLineChartDraws(t *testing.T) {
...
+		{
+			desc:   "custom Y-axis labels using a value formatter that returns empty labels",
+			canvas: image.Rect(0, 0, 20, 10),
+			opts: []Option{
+				YAxisFormattedValues(func(v float64) string { return "" }),
+			},
+			writes: func(lc *LineChart) error {
+				return lc.Series("first", []float64{0, 100})
+			},
+			wantDrawErr: true,
```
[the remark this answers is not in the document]
slok: Yes, sometimes is useful to remove the Y-axis, so the reader focuses on the visual pattern.

If is something that is not complex I'll make it in this PR, if it takes time I think that is better to have this feature (the formatter) and add erase the Y-axis later, in the end, this feature wasn't available neither, so in any case, we are winning product value :)

slok:
I pushed the changes to allow empty labels on the Y-axis

mum4k:
Wow that was fast, looks good. Thank you.

widgets/linechart/linechart_test.go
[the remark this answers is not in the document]
mum4k: Agreed and glad to see that it turned out to be easy.

slok:
I've pushed the code for a single unit time duration value formatter.

It's a factory method that returns a value formatter that It can be customized with the number of decimals and the time unit.

I've added also a helper for second based 0 decimals value formatter, that is a common case for RPC based services like GRPC or HTTP latency graphs.

What do you think?

widgets/linechart/formatter.go
```diff
@@ -0,0 +1,85 @@
...
+	prefix := ""
+	if d < 0 {
+		prefix = "-"
+		d = time.Duration(math.Abs(d.Seconds()) * float64(time.Second))
+	}
+
+	switch {
+	// Nanoseconds.
+	case d.Nanoseconds() < 1000:
+		dfmt := prefix + "%dns"
```
mum4k: This switch reads really well, thanks for using such a well structured approach.

(nit) Go tends to use mixedCaps for multi-word variable names.

Can we call this variable dFmt ?

https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps

widgets/linechart/formatter_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"time"
+
+	"github.com/kylelemons/godebug/pretty"
+)
+
+func TestFormatters(t *testing.T) {
+	tests := []struct {
+		desc      string
+		value     float64
+		formatter func(float64) string
```
mum4k: (optional) We could reference our type here which will serve as documentation. Test cases are often used when discovering how code works.

```go
formatter ValueFormatter
```

widgets/linechart/formatter.go
```diff
@@ -0,0 +1,85 @@
...
+// durationSingleUnitPrettyFormat returns the pretty format in one single
+// unit for a time.Duration, the different returned unit formats
+// are: nanoseconds, microseconds, milliseconds, seconds, minutes
+// hours, days.
+func durationSingleUnitPrettyFormat(d time.Duration, decimals int) string {
+	// Check if the duration is less than 0.
+	prefix := ""
+	if d < 0 {
+		prefix = "-"
+		d = time.Duration(math.Abs(d.Seconds()) * float64(time.Second))
```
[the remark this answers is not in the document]
slok: It works ok (https://play.golang.org/p/o2nfSMCMS3A). I'll add a test!

widgets/linechart/formatter.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func decimalFormat(decimals int, unit string) string {
+	return fmt.Sprintf("%%.%df%s", decimals, unit)
+}
+
+// SingleUnitDurationValueFormatter is a factory to create a custom duration
+// in a single unit representation formatter based on a unit and the decimals
+// to truncate.
+func SingleUnitDurationValueFormatter(unit time.Duration, decimals int) ValueFormatter {
```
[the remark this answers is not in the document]
slok: Better, yes

widgets/linechart/formatter.go
```diff
@@ -0,0 +1,85 @@
...
+		d := time.Duration(v * float64(unit))
+		return durationSingleUnitPrettyFormat(d, decimals)
+	}
+}
+
+// SingleUnitSecondsValueFormatter is a formatter that will receive
+// seconds unit in the float64 argument and will return a pretty
+// format in one single unit without decimals, it doesn't round,
+// it truncates.
+var SingleUnitSecondsValueFormatter = SingleUnitDurationValueFormatter(time.Second, 0)
```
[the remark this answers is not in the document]
slok: I'm against global state and try injecting every dependency, but I thought that this would be handy for users and like sentinel errors (`io.EOF`) it should respect and not alter that variable..., although I'm not against changing and making more secure the declaration because as you say the user always can do crazy things..., thanks!

widgets/linechart/formatter.go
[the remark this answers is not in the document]
slok: You are right (as always haha!), thanks

widgets/linechart/formatter.go
slok: In that case, shouldn't be `dFMT`? (https://github.com/golang/go/wiki/CodeReviewComments#initialisms)

slok:
Ready for a new check.

I've also added the check for `NaN` values.

Thanks @mum4k!

widgets/linechart/formatter.go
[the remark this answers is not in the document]
mum4k: Ah of course, it works because duration.Seconds() returns a float that can represent sub-second values. I didn't realize that. Thank you for the proof and the unit test.

widgets/linechart/formatter.go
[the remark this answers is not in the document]
mum4k: Exactly, you summarized it well. This remind me of an xkcd episode actually, which I am sure you saw before:

https://xkcd.com/1172/

widgets/linechart/formatter.go
mum4k: I was actually quite confused between the two for a while. The way I understand it, fmt is an abbreviation, not an initialism. Abbreviations (the word "format" shortened to fmt) follow the standard mixedCaps rules.

Initialisms (each letter stands for its own word, like HTTP) use the second rule.

Either way this is mostly informational as these things are a bit too much on the "extreme style police" side of code reviews.

mum4k:
Looks good @slok.

Good catch about checking for the NaN values, I completely missed that.

widgets/linechart/formatter.go
slok: Right! I've always been a little bit confused with this, anyway `dFmt` it's perfect :)

slok:
I pushed the last step. Add a simple round formatter.

I've added also suffix and decimal based formatter that I think that can be helpful for the users in lots of simple cases where you only want to add a suffix and limit the decimals so it shows pretty on the line chart.

I've also updated the changelog.

When we are finished with this last step I'll remove the `draft` state from the PR and would be ready :tada:

widgets/linechart/value_formatter.go
```diff
@@ -42,33 +43,34 @@ func durationSingleUnitPrettyFormat(d time.Duration, decimals int) string {
...
-		dFmt := prefix + decimalFormat(decimals, "d")
+		dFmt := prefix + suffixDecimalFormat(decimals, "d")
 		return fmt.Sprintf(dFmt, d.Hours()/24)
 	}
 }
 
-func decimalFormat(decimals int, unit string) string {
-	return fmt.Sprintf("%%.%df%s", decimals, unit)
+func suffixDecimalFormat(decimals int, suffix string) string {
+	suffix = strings.ReplaceAll(suffix, "%", "%%") // safe `%` character for fmt.
```
mum4k: Another great catch, thanks for being so thorough @slok!

CHANGELOG.md
```diff
@@ -9,9 +9,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- The `LineChart` can format the labels on the Y-axis with a `ValueFormatter`.
 - The `SegmentDisplay` can now display dots and colons ('.' and ':').
 - The `Donut` widget now guarantees spacing between the donut and its label.
 
+### Added
```
mum4k: (nit) Would you mind moving Added before Changed?

slok:
Ready! :)

mum4k:
For some reason the automated checks didn't trigger. I will try to cause a Travis run by closing and reopening this PR.

widgets/linechart/value_formatter.go
```diff
@@ -0,0 +1,146 @@
...
+		return fmt.Sprintf(dFmt, d.Hours())
+	// Days.
+	default:
+		dFmt := prefix + suffixDecimalFormat(decimals, "d")
+		return fmt.Sprintf(dFmt, d.Hours()/24)
+	}
+}
+
+func suffixDecimalFormat(decimals int, suffix string) string {
+	suffix = strings.ReplaceAll(suffix, "%", "%%") // safe `%` character for fmt.
```
mum4k: Looks like this was added in Go 1.12. We are trying to support the last three versions which currently includes Go 1.10 and 1.11.

https://travis-ci.org/mum4k/termdash/builds/529547980

Can we find a solution compatible with the older Go versions? One option (what I have done in the past) is to simply copy the implementation of the missing strings.ReplaceAll into the Termdash repo with a note that it can be removed when we drop support for Go 1.10 and 1.11.

We could create our own strings package under the internal directory:

https://github.com/mum4k/termdash/tree/master/internal

slok: After all the changes I broke the compatibility In the last change haha!

I think that we can fallback to the old known `Replace(s, old, new, -1)`, that is what it does the `ReplaceAll` internally. I don't think that is worth creating a new package for this simple statement. I'll push this change for now and if you prefer to have the internal package I'll do the change.

slok:
I forgot about the permission, sorry!

You have permission to merge this PR into `master` branch.

widgets/linechart/value_formatter.go
mum4k: Of course, that is much better. Thank you.

mum4k:
Looks good, thanks a lot @slok, I have really enjoyed working with you on this feature.

slok:
Same here! Thanks for your time and effort in explaining things. I hope to contribute more!

mum4k:
Anytime, looking forward to it already. How is your project going, are you approaching a release? Are there any other major features you are missing?

slok:
I'm implementing the unit formatters inside the app at this moment (maybe we could port some of them to `termdash`), when finished, next will be a little bit of documentation, clean up the readme, add some flags that I think that will be handy and I'll be ready for the first release, depending on how many hours I can spend this weekend I think that next week I'll have it!

At this moment for the first release of `grafterm` I think that the required ones are already implemented, if people like this app and start asking for features I'll have a better priority vision.

In mind, I have lot's of features apart from the ones we discussed in the other PR (the table and the selector are the ones that I think are the most important ones). Maybe are a little crazy xD but I give you some of them in case you like or you think that will be fun to implement.

- Stacked graph series, this can be done calculating the final value aggregating the different values by the users of `termdash`, but using a simple flag on the graph widget would be the easiest way for the user (I don't know if this is worth).
- Legend over a graph (I don't know if this is possible)
- Fill the graph lines area (render all area instead of only a line, this could come with Z-axis score/value, to set the depth of series so the known smallest ones can be over the highest ones, this way they are not hidden, e.g. percentiles).

Example all these on the same graph in Grafana



- [Heatmap](https://grafana.com/docs/features/panels/heatmap/#heatmap) widget.
- A more feature-rich bar gauge (check the ideas [here](https://play.grafana.org/d/ZvPm55mWk/new-features-in-v6-2?orgId=1)).
  - horizontal, vertical.
  - Multiple colors on the bar based on the bar percent.

Like I said, some maybe are crazy ideas :)

I'll notify you when [Grafterm](https://github.com/slok/grafterm) is released!

mum4k:
That sounds great @slok, looking forward to the first release of *grafterm*. Feel free to circle back and open issues if you find anything wrong or missing. Meanwhile I will continue working on the table and the selector, same as you it is a continuous fight in trying to find enough hours during the weekend.

No ideas are crazy and brainstorming really helps, so feel free to propose anything that comes to mind, worst case we will agree that it isn't doable or worth it ;)

The **stacked graph series** and **filled graph areas** look really good, but might not be doable on terminal due to limitations that I didn't find a way around yet. Even though with the braille unicode characters we get sub-character resolution, each cell is still limited to only one set of cell options. This becomes a problem especially when two series share a cell, the entire cell would be of a single color (series 1 or series2). You can already observe this in cells where multiple series cross, when we start filling them it might look bad or the differences / borders between series might get lost. If you have ideas how to overcome this - please let me know.

The **legend** over a graph is absolutely doable once we have #104. Same as above, we can only let users click on a full cell, rather than a braille point within a cell, but that shouldn't be a large issue. If you have some spare time - please do file an issue briefly describing the functionality (what should be displayed in the pop-up).

The **heatmap** might also suffer from the single-color-per-cell issue, unless we create a cell based heatmap which might have too low of a resolution? What do you think?

**More features for bar chart** sound good, I was already planning a #9. As with the legend - if you get some spare cycles, please file an issue describing what features we should add.

Thanks for all the ideas and support!

slok:
Hey @mum4k, I've been lucky and manage to get the time and have a productive weekend :tada: https://github.com/slok/grafterm

mum4k:
Looks amazing, great work and congratulations @slok. Glad to see this come together.

I took this opportunity and listed **grafterm** in the [README](https://github.com/mum4k/termdash/commit/ce095e1447ef1581ea9988cdb1f5cf32719c879c#diff-04c6e90faac2675aa89e2176d2eec7d8), however if you would prefer I don't, please let me know.

Alternatively feel free to send a PR against master adding a short description next to the link if you would like.

slok:
Many thanks for adding a link to grafterm on the readme!

PR #195, opened 2019-04-29
Release v0.9.0

mum4k:
### Added

- The `TextInput` widget, an input field allowing interactive text input.
- The `Donut` widget can now display an optional text label under the donut.

### Changed

- Widgets now get information whether their container is focused when Draw is
  executed.
- The SegmentDisplay widget now has a method that returns the observed character
  capacity the last time Draw was called.
- The grid.Builder API now allows users to specify options for intermediate
  containers, i.e. containers that don't have widgets, but represent rows and
  columns.
- Line chart widget now allows `math.NaN` values to represent "no value" (values
  that will not be rendered) in the values slice.

#### Breaking API changes

- The widgetapi.Widget.Draw method now accepts a second argument which provides
  widgets with additional metadata. This affects all implemented widgets.
- Termdash now requires at least Go version 1.10, which allows us to utilize
  `math.Round` instead of our own implementation and `strings.Builder` instead
  of `bytes.Buffer`.
- Terminal shortcuts like `Ctrl-A` no longer come as two separate events,
  Termdash now mirrors termbox-go and sends these as one event.


mum4k:
@slok confirmed and gave permission to push his code into the master branch in #185.

PR #203, opened 2019-05-04
Implementing the Table widget.

### LINKED ISSUE #5: Widget: Table
Implement a text table widget.

Similar to: https://github.com/yaronn/blessed-contrib/blob/master/docs/images/table.gif

Targeted features:

- Support both ASCII and Unicode text.
- API to specify columns and rows.
- API to specify sizes of columns (auto based on content, fixed or percentage).
- API to merge multiple columns and rows (colspan and rowspan).
- Support content alignment within cells.
- Support Text trimming, character or word wrapping within cells.
- Support scrolling of longer tables.
- Support keyboard and mouse input to interact with the table (row by row or cell by cell).
- Visible or invisible cell and table borders.
- Support sorting by individual columns (ascending and descending).

mum4k:
Fixes #5.

Steps:
- [x] Define the API.
- [x] Calculate column widths dynamically according to content.
- [x] Account for padding in algorithm that calculates column widths.
- [x] Given cvsWidth, calculate width available to cells (minus border and padding).
- [ ] Each row should report its required height (cell content + padding).
- [ ] Draw row by row, bottom up or top-down (direction of scroll). Support partially drawn rows.
- [ ] Function that returns an area available for a cell.
- [ ] Function on each cell that draws its content (best effort on size at this point).
- [ ] Ensure content copies data given by the user to prevent races.
- [ ] Support sorting by columns.


kvnxiao:
Is this widget still being worked on? Anything blocking it from being merged?

mum4k:
This is only about 30% done. We have the API and a bit that dynamically chooses column widths. None of the rendering bits exists as of yet.

It is not being actively worked on at the moment and I won't get to it in the immediate future. I am planning to finish it in the long term. Alternatively if someone else wants to pick this up, I can merge the bits that are done. Or we can start on a clean slate depending on preference.

mum4k:
Closing for inactivity.

PR #204, opened 2019-05-05
Add support for fixed size container splits.

nijynot:
This PR adds support for setting fixed heights to containers that split horizontally.

---

**Examples:** Calling
```
container.SplitHorizontal(
	container.Top(...),
	container.Bottom(...),
	container.FixHeight(3, 15)
)
```
will set the top container to be `3` in line height and `15` on the bottom container.

To make the containers fill the rest of the space, you set the args to `-1`, example:
```
container.SplitHorizontal(
	container.Top(...),
	container.Bottom(...),
	container.FixHeight(3, -1)
)
```
which will make top container `3` and make the bottom container take up rest of the space that is available.

### DIFF
--- a/container/container_test.go
+++ b/container/container_test.go
@@ -523,7 +523,7 @@ func TestNew(t *testing.T) {
 						Bottom(
 							Border(linestyle.Light),
 						),
-						SplitPercent(-4),
+						SplitFixed(-2),
 					),
 				)
 			},
--- a/container/container_test.go
+++ b/container/container_test.go
@@ -491,6 +491,26 @@ func TestNew(t *testing.T) {
 			wantContainerErr: true,
 		},
 		{
+			desc:     "fails when both SplitFixed and SplitPercent are specified",
+			termSize: image.Point{10, 10},
+			container: func(ft *faketerm.Terminal) (*Container, error) {
+				return New(
+					ft,
+					SplitHorizontal(
+						Top(
+							Border(linestyle.Light),
+						),
+						Bottom(
+							Border(linestyle.Light),
+						),
+						SplitFixed(4),
+						SplitPercent(20),
+					),
+				)
+			},
+			wantContainerErr: true,
+		},
+		{
 			desc:     "empty container",
 			termSize: image.Point{10, 10},
 			container: func(ft *faketerm.Terminal) (*Container, error) {
@@ -617,6 +637,32 @@ func TestNew(t *testing.T) {
 			},
 		},
 		{
+			desc:     "horizontal unequal split",
+			termSize: image.Point{10, 20},
+			container: func(ft *faketerm.Terminal) (*Container, error) {
+				return New(
+					ft,
+					SplitHorizontal(
+						Top(
+							Border(linestyle.Light),
+						),
+						Bottom(
+							Border(linestyle.Light),
+						),
+						SplitFixed(4),
+					),
+				)
+			},
+			want: func(size image.Point) *faketerm.Terminal {
+				ft := faketerm.MustNew(size)
+				cvs := testcanvas.MustNew(ft.Area())
+				testdraw.MustBorder(cvs, image.Rect(0, 0, 10, 4))
+				testdraw.MustBorder(cvs, image.Rect(0, 4, 10, 20))
+				testcanvas.MustApply(cvs, ft)
+				return ft
+			},
+		},
+		{
 			desc:     "horizontal split, parent and children have borders",
 			termSize: image.Point{10, 10},
 			container: func(ft *faketerm.Terminal) (*Container, error) {
@@ -730,6 +776,32 @@ func TestNew(t *testing.T) {
 							Border(linestyle.Light),
 						),
 						SplitPercent(20),
+					),
+				)
+			},
+			want: func(size image.Point) *faketerm.Terminal {
+				ft := faketerm.MustNew(size)
+				cvs := testcanvas.MustNew(ft.Area())
+				testdraw.MustBorder(cvs, image.Rect(0, 0, 4, 10))
+				testdraw.MustBorder(cvs, image.Rect(4, 0, 20, 10))
+				testcanvas.MustApply(cvs, ft)
+				return ft
+			},
+		},
+		{
+			desc:     "vertical fixed splits",
+			termSize: image.Point{20, 10},
+			container: func(ft *faketerm.Terminal) (*Container, error) {
+				return New(
+					ft,
+					SplitVertical(
+						Left(
+							Border(linestyle.Light),
+						),
+						Right(
+							Border(linestyle.Light),
+						),
+						SplitFixed(4),
 					),
 				)
 			},
--- a/container/container_test.go
+++ b/container/container_test.go
@@ -511,6 +511,25 @@ func TestNew(t *testing.T) {
 			wantContainerErr: true,
 		},
 		{
+			desc:     "fails on SplitFixed less than -1",
+			termSize: image.Point{10, 20},
+			container: func(ft *faketerm.Terminal) (*Container, error) {
+				return New(
+					ft,
+					SplitHorizontal(
+						Top(
+							Border(linestyle.Light),
+						),
+						Bottom(
+							Border(linestyle.Light),
+						),
+						SplitPercent(-4),
+					),
+				)
+			},
+			wantContainerErr: true,
+		},
+		{
 			desc:     "empty container",
 			termSize: image.Point{10, 10},
 			container: func(ft *faketerm.Terminal) (*Container, error) {
--- a/container/options.go
+++ b/container/options.go
@@ -52,7 +52,7 @@ func validateIds(c *Container, seen map[string]bool) error {
 
 // ensure all the container only have one split modifier.
 func validateSplits(c *Container) error {
-	if c.opts.splitFixed > -1 && c.opts.splitPercent != 50 {
+	if c.opts.splitFixed > DefaultSplitFixed && c.opts.splitPercent != DefaultSplitPercent {
 		return fmt.Errorf(
 			"only one of splitFixed `%v` and splitPercent `%v` is allowed to be set per container",
 			c.opts.splitFixed,
@@ -245,13 +245,15 @@ func SplitPercent(p int) SplitOption {
 	})
 }
 
-// SplitFixed sets the size as a fixed value of the available space.
+// SplitFixed sets the size of the first container to be a fixed value
+// and makes the second container take up the remaining space.
 // When using SplitVertical, the provided size is applied to the new left
 // container, the new right container gets the reminder of the size.
 // When using SplitHorizontal, the provided size is applied to the new top
 // container, the new bottom container gets the reminder of the size.
 // The provided value must be a positive number in the range 0 <= cells.
-// If not provided, defaults to DefaultSplitFixed.
+// If SplitFixed() is not specified, it defaults to SplitPercent() and its given value.
+// Only one of SplitFixed() and SplitPercent() can be specified per container.
 func SplitFixed(cells int) SplitOption {
 	return splitOption(func(opts *options) error {
 		if cells < 0 {
--- a/container/options.go
+++ b/container/options.go
@@ -77,6 +77,8 @@ type options struct {
 	// split identifies how is this container split.
 	split        splitType
 	splitPercent int
+	topHeight int
+	bottomHeight int
 
 	// widget is the widget in the container.
 	// A container can have either two sub containers (left and right) or a
@@ -167,6 +169,8 @@ func newOptions(parent *options) *options {
 		hAlign:       align.HorizontalCenter,
 		vAlign:       align.VerticalMiddle,
 		splitPercent: DefaultSplitPercent,
+		topHeight: -1,
+		bottomHeight: -1,
 	}
 	if parent != nil {
 		opts.inherited = parent.inherited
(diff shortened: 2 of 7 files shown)


container/options.go
```diff
@@ -216,6 +220,20 @@ func SplitPercent(p int) SplitOption {
 	})
 }
 
+// FixHeight sets the heights of the top and the bottom containers.
+// If an argument is equal to -1, then the height will be set automatically,
+// i.e. will fill the rest of the space that is available.
+func FixHeight(th int, bh int) SplitOption {
```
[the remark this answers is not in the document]
nijynot: Sounds good to me. :ok_hand:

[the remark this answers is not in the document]
nijynot: Ah, gotcha. Will fix :+1:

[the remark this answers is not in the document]
nijynot: It takes two integers as you can then stack containers without having unreachable cells.

Example:
```
c, err := container.New(
  t,
  container.SplitHorizontal(
    container.Top(
      container.Border(linestyle.Light),
      container.PlaceWidget(test),
    ),
    container.Bottom(
      container.SplitHorizontal(
        container.Top(
          container.Border(linestyle.Light),
          container.PlaceWidget(test),
        ),
        container.Bottom(
          container.SplitHorizontal(
            container.Top(
              container.Border(linestyle.Light),
              container.PlaceWidget(test),
            ),
            container.Bottom(
              container.Border(linestyle.Light),
              container.PlaceWidget(test),
            ),
            container.FixHeight(3, -1),
          ),
        ),
        container.FixHeight(3, -1),
      ),
    ),
    container.FixHeight(3, -1),
  ),
)
```
Code above will give us:

Screenshot from 2019-05-05 23-28-43

where the first 3 widgets are fixed. While if `SplitFixed` only takes one argument, this would not be possible I think?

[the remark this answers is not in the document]
mum4k: Thank you for the example code and the screenshot, this helps a lot. I think we might be talking about the same thing. We aren't really using the second argument in your example, it is always set to -1.

If container.SplitFixed() takes just one argument (the number of cells for the top container on a horizontal split or the number of cells for the left container on a vertical split) the above code should work in exactly the same way? Unless of course I misunderstood something.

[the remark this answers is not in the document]
nijynot: Oh yeah, we're talking about the same thing.

Back to your previous comment on
> Are there any use cases that wouldn't be satisfying with such an option? It might be better if we focus on those and see if we really need two arguments.

Makes sense, having two arguments probably won't have much use case as most will want to fill the space. Will change it to `SplitFixed(cells int)` instead.

container/options.go
```diff
@@ -216,6 +218,19 @@ func SplitPercent(p int) SplitOption {
 	})
 }
 
+// SplitFixed sets the heights of the top and the bottom containers.
```
mum4k: We should probably update this comment so that it no longer references "top" or "bottom". The way we have it now, it works on either horizontal or vertical splits.

Maybe we can borrow some language from the comment above SplitPercent?

container/options.go
```diff
@@ -216,6 +218,19 @@ func SplitPercent(p int) SplitOption {
...
+// The first container will be fixed and the second one will fill the rest of the space.
+// Allows values greater or equal to zero.
+func SplitFixed(cells int) SplitOption {
+	return splitOption(func(opts *options) error {
+		if cells < 0 {
```
mum4k: I see that SplitPercent does the same thing, but it probably is worth asking: Is it valid to specify SplitFixed(0) ? Or should it be an error.

Setting it to zero is the same thing as not performing the split at all. What do you think?

container/options.go
```diff
@@ -216,6 +218,19 @@ func SplitPercent(p int) SplitOption {
...
+			return fmt.Errorf("invalid fixed value %d, must be in range %d <= cells", cells, 0)
```
mum4k: Now that we have two ways of specifying split size, we should validate that only one of them is ever set. I.e. return an error if both are set.

Such validation can be added into this function:

https://github.com/mum4k/termdash/blob/0e5242ca6f8386889a303c67b58a25f98aed75b6/container/options.go#L41-L61

internal/area/area.go
```diff
@@ -42,16 +42,29 @@ func FromSize(size image.Point) (image.Rectangle, error) {
 // specified percentage of its width. The percentage must be in the range
 // 0 <= heightPerc <= 100.
 // Can return zero size areas.
-func HSplit(area image.Rectangle, heightPerc int) (top image.Rectangle, bottom image.Rectangle, err error) {
+//
+// The fixed value must by in the range -1 <= heightFixed.
+// If fixed value is -1 (default value), heightPerc is used instead.
+func HSplit(area image.Rectangle, heightPerc int, heightFixed int) (top image.Rectangle, bottom image.Rectangle, err error) {
```
mum4k: One of good principles to follow when it comes to code readability is to keep functions focused on just one thing. The HSplit() and VSplit() functions perform relative splits based on percentage. Adding another argument that changes their behavior makes both the calling code and these functions harder to understand.

Normally the approach would be to implement two new functions that perform fixed splits, however it looks like we are lucky and those already exist:

https://github.com/mum4k/termdash/blob/0e5242ca6f8386889a303c67b58a25f98aed75b6/internal/area/area.go#L81-L86

https://github.com/mum4k/termdash/blob/0e5242ca6f8386889a303c67b58a25f98aed75b6/internal/area/area.go#L104-L109

Would you mind taking a look if they can be used out of the box or if they need to be modified? Once they are ready for us, we can modify the calling code to choose which function to call based on which option was provided.

mum4k:
(fyi) I have changed the target branch to devel. We try to follow this branching model with Termdash:

https://nvie.com/posts/a-successful-git-branching-model/

New features get developed and pushed to devel and then released once a month into master.

container/options.go
nijynot: I think we should support SplitFixed(0) and SplitPercent(0).

It doesn't seem to hurt to support 0, and it maybe can be used to hide and show the container.
I.e., instead of unmounting and mounting a container when wanting to show and hide, you can maybe just set the value to zero and then set it to bigger than zero.

So supporting zero might have some functionality.

mum4k: That is a good point, thank you for providing the example.

Let's keep supporting zero as you suggested.

container/options.go
nijynot: Split the `validateOptions` into two sub functions -- `validateIds` and `validateSplits`.
`validateIds` checks that no duplicate ids exists, and `validateSplits` check that SplitFixed and SplitPercent are not specified at the same time.

internal/area/area.go
nijynot: These worked great. Dropped them in and it just works

mum4k:
(fyi) The CLA issue appeared because the devel was behind master due to some commits from other authors that were pushed directly. I have merged them back to devel and will attempt to rebase this PR so that we only see your commits.

mum4k:
CLA issue resolved manually, this PR now only includes commits from @nijynot.

container/options.go
mum4k: It is fine to split these into two functions, more focused functions are easier to digest. We should however consider the cost of the preOrder traversal. The containers are organized in a binary tree and now each of the validation functions traverses from the root container to all the leaves, visiting and validating each container.

While we probably aren't going to feel the performance impact immediately, it might be a low hanging fruit to perform the traversal just once and call the two validation functions on each visited container.

Do you think this is something worth fixing?

container/container_test.go
```diff
@@ -490,6 +490,26 @@ func TestNew(t *testing.T) {
 			},
 			wantContainerErr: true,
 		},
+		{
+			desc:     "fails when both SplitFixed and SplitPercent are specified",
```
mum4k: Sorry if commenting prematurely, not sure if you're still working on this.

Thank you for adding the test coverage for the new feature - can we also add a test case when the user specifies invalid SplitFixed value, like SplitFixed(-4) ?

The condition seems to be missing coverage:
https://coveralls.io/builds/23268685/source?filename=container%2Foptions.go#L263

container/options.go
```diff
@@ -60,6 +59,43 @@ func validateOptions(c *Container) error {
...
+	if errStr != "" {
+		return errors.New(errStr)
+	}
+	return nil
+}
+
+// validateOptions validates options set in the container tree.
+func validateOptions(c *Container) error {
+	// ensure all the container identifiers are either empty or unique.
+	errIds := validateIds(c)
```
mum4k: (Go style nit, optional) We usually call the error variable of type error just "err". We should also try to minimize the scope of the error variable, that way future error checks below won't shadow the variable.

A Go idiomatic error check would look like this:
if err := validateIds(c); err != nil {
  return err
}

Notice that the scope of the "err" variable is container within the if statement where it was declared.

mum4k: (optional) If we want to we could reduce the multiple traversal issue by calling preOrder just once from within validateOptions().

Related to an earlier comment.

internal/area/area.go
```diff
@@ -42,16 +42,20 @@ func FromSize(size image.Point) (image.Rectangle, error) {
 // specified percentage of its width. The percentage must be in the range
 // 0 <= heightPerc <= 100.
 // Can return zero size areas.
+//
```
mum4k: Do we want to revert the changes done to the area.go file?

container/container_test.go
nijynot: Not at all, you commented at the right point.

Will cover that condition. :+1:

internal/area/area.go
nijynot: Good catch, reverted it.

container/options.go
nijynot: Makes sense. But yeah, as you said, no big impact atm, although fixed it anyways.

container/container_test.go
```diff
@@ -510,6 +510,25 @@ func TestNew(t *testing.T) {
...
+				return New(
+					ft,
+					SplitHorizontal(
+						Top(
+							Border(linestyle.Light),
+						),
+						Bottom(
+							Border(linestyle.Light),
+						),
+						SplitPercent(-4),
```
mum4k: Did we want to use SplitFixed() here?

(optional) As for the chosen value, it is better to choose the "edge value", i.e. the first one that would be incorrect (-2). If we choose -4, we aren't really covering -3 and -2 with this test.

container/options.go
```diff
@@ -38,25 +38,49 @@ func applyOptions(c *Container, opts ...Option) error {
 	return nil
 }
 
+// ensure all the container identifiers are either empty or unique.
+func validateIds(c *Container, seen *(map[string]bool)) error {
```
mum4k: Do we need to accept a pointer to the map here?

Maps are actually reference types in Go, they are like a pointer by themselves so this should just work. Here is a generic advice on when to use a pointer:

https://github.com/golang/go/wiki/CodeReviewComments#receiver-type

container/options.go
```diff
@@ -216,6 +242,19 @@ func SplitPercent(p int) SplitOption {
 	})
 }
 
+// SplitFixed sets the size as a fixed value of the available space.
+// FIXME: More comments
```
mum4k: Did we want to remove the FIXME and add comments? We could take what was written for SplitPercent and adjust accordingly.

container/container_test.go
nijynot: Oops, you're right, should be SplitFixed here.
-2 is a better value too, agreed, will fix.

container/options.go
nijynot: Thanks for the link, very helpful.
Will fix.

container/options.go
nijynot: Forgot about this one, was a comment for myself before I pushed changes, but slipped my mind.
I'll add comments that are similar to SplitPercent. :+1:

nijynot:
Addressed the comments, and also added a `const DefaultSplitFixed = -1` similar to `const DefaultSplitPercent = 50`.

container/options.go
```diff
@@ -38,25 +38,49 @@ func applyOptions(c *Container, opts ...Option) error {
...
+		return fmt.Errorf("duplicate container ID %q", c.opts.id)
+	}
+	seen[c.opts.id] = true
+
+	return nil
+}
+
+// ensure all the container only have one split modifier.
+func validateSplits(c *Container) error {
+	if c.opts.splitFixed > -1 && c.opts.splitPercent != 50 {
```
mum4k: (optional) We might want to use DefaultSplitPercent here instead of the literal 50. That way if anyone ever decides to change the value - they don't forget to change it here too.

https://github.com/mum4k/termdash/blob/0e5242ca6f8386889a303c67b58a25f98aed75b6/container/options.go#L200

Optional, it is unlikely we will be changing the default behavior as that would be a breaking change for multiple users.

container/options.go
```diff
@@ -216,6 +245,23 @@ func SplitPercent(p int) SplitOption {
 	})
 }
 
+// SplitFixed sets the size as a fixed value of the available space.
```
mum4k: (nit) It is a bit unclear which size we are talking about. Can we say:

// SplitFixed sets the relative size of the split to a fixed number of cells.

Or similar.

container/options.go
```diff
@@ -216,6 +245,23 @@ func SplitPercent(p int) SplitOption {
...
+// When using SplitVertical, the provided size is applied to the new left
+// container, the new right container gets the reminder of the size.
+// When using SplitHorizontal, the provided size is applied to the new top
+// container, the new bottom container gets the reminder of the size.
+// The provided value must be a positive number in the range 0 <= cells.
+// If not provided, defaults to DefaultSplitFixed.
```
mum4k: Can we also add a sentence here (and in the comment for SplitPercent) saying that only one can be used?

Something like:

// Only one of SplitFixed() and SplitPercent() can be specified.

mum4k: The note about the "default" might not be correct. Since when this isn't provided, this really defaults to using SplitPercent with its default value of  DefaultSplitPercent.

container/options.go
nijynot: Did the change and changed the literal -1 to DefaultSplitFixed too. :ok_hand:

container/options.go
nijynot: Went with:
```
// SplitFixed sets the size of the first container to be a fixed value
// and makes the second container take up the remaining space.
```
A bit longer, but a bit easier to understand imo. Is this ok or should I change it to
`// SplitFixed sets the relative size of the split to a fixed number of cells.`?

container/options.go
nijynot: Ah, yeah that's more accurate.
Changed to
```
// If SplitFixed() is not specified, it defaults to SplitPercent() and its given value.
```

nijynot:
No worries, rather appreciate your comments!
And you have my permission to push this code into master branch at the next release. :+1:

PR #209, opened 2019-05-19
Added indicator widget (tests pending)

### LINKED ISSUE #216: Widget: Indicator
Implement a widget that can indicate activity, like an HDD or modem light.

keithknott26:
This PR Serves to add the indicator branch as well as add the basic indicator functionality. I'd appreciate your comments on whether or not this widget should draw the indicator in a different way. Currently its using a unicode character to designate on/off.  Another idea might be to have a rectangle filled with a color representing on/off?

Please run the indicatordemo to see the functionality, it supports text labels underneath similar to the donut however I'm having difficulty getting the text to align properly on resize. It may be just that the container is resizing.

Fixes #216


keithknott26:
> I signed it!

I signed it!

widgets/indicator/indicator.go
```diff
@@ -0,0 +1,206 @@
...
+	} else {
+		indAr = cvs.Area()
+	}
+	var t string
+	// split the indicator and label
+	// checked: \u25C9
+	// unchecked: \u25EF
+	// line: \u2500─────────────
+	if i.status == true {
+		t = "\u25C9"
```
mum4k: The next thing we should talk about is scaling. Most of the Termdash widgets scale in size as the container grows. If we just implement the drawing as a single character, it will always have a fixed size.

This is really up to you as the author of the widget. Do you think we should have this as a fixed size widget (the current implementation)? Or should we use the character only when the size of the container is small and use two concentric circles on a braille canvas as the size grows?

widgets/indicator/indicator.go
```diff
@@ -0,0 +1,206 @@
...
+func (i *Indicator) Flash() error {
+	if i.status == true {
+		i.status = false
+	} else {
+		i.status = true
+	}
+	return nil
+}
+
+func (i *Indicator) Pulse() error {
```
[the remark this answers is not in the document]
keithknott26: @mum4k, 

I ran into this exact scenario when creating the demo. I really like your approach to problems and development, you write neat, concise, readable code. Having On, Off, and Toggle as you described makes the most sense I agree.

widgets/indicator/indicator.go
keithknott26: You raise a really good point, I really like your idea of scaling up to be a braille canvas widget and scaling down to be a unicode character... I'll give that a shot and see if it looks ok.

mum4k: Sounds good, thanks @keithknott26.

To be noted - we can also agree to add the "scaling" at a later time, it isn't a feature that has to be in from day one. Completely up to you.

keithknott26: When you have a moment , have a look at the indicator demo and let me know if you think that looks ok. If so I'll try to figure out how to display the scaled version and then the unicode version when things get below 3,3

keithknott26: Also, I was hoping I to apply some mouse handling to this indicator as well since it now has the on/off/toggle methods. I'd like your suggestion on implementing a type of "onclick" event handler, it might be beyond my capabilities but I think it would be a good addition to this widget

widgets/indicator/options.go
```diff
@@ -0,0 +1,97 @@
...
+			cell.FgColor(cell.ColorRed),
+			cell.BgColor(cell.ColorDefault),
+		},
+		labelAlign: DefaultLabelAlign,
+	}
+}
+
+// TextCellOpts sets cell options on cells that contain the displayed text
+// progress.
+func TextCellOpts(cOpts ...cell.Option) Option {
```
mum4k: Does it make sense to pass in cell options here? They allow to pass both foreground and background colors, but the indicator isn't always a text character anymore.

Should we rename this to just Color? I.e. an option that sets the color of the indicator?

Looking below we also have CellOpts. It isn't very clear which one is for what and the text on them still mentions the Donut widget (probably and copy&paste).

Can we try to cleanup this options file before we move on with the review since this is also part of the public API.

widgets/indicator/indicator.go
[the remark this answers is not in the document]
mum4k: Thanks Keith, the public API looks good now.

widgets/indicator/indicator.go
mum4k: I have checked out the demo and it looks good! The scaling looks good.. We should probably add logic that chooses one or the other based on the canvas size, i.e. when it is below {3, 3} we should only use the character version, otherwise we should use the scaled version.

Secondly we might want to add an Option that sets the maximum indicator size as a diameter and when 1-3 is passed, it will only use the character version. So that the user can limit the scaling. What do you think?

When it comes to handling mouse events - that actually isn't too hard. There is an implementation of a mouse state machine (tracks clicks and determines when a full click happened). A full click is when the button is both pressed and released in the widget's area. You can look at the button widget and how the FSM is used. Before we go there - what do you envision this used as? What should happen when the user clicks on the indicator?

widgets/indicator/indicator.go
```diff
@@ -0,0 +1,244 @@
...
+	r := int(math.Round(float64(indicatorRadius) / 100 * float64(90)))
+	if r < 2 { // Smallest possible circle radius.
+		return 0
+	}
+	return r
+}
+
+// Draw draws the Indicator widget onto the canvas.
+// Implements widgetapi.Widget.Draw.
+func (i *Indicator) Draw(cvs *canvas.Canvas, meta *widgetapi.Meta) error {
```
mum4k: One more thing to consider - the text label is fairly close to the indicator. We might want to consider leaving at least one empty line between them, similarly to what the Donut widget does. This was changed in the following commit for reference.

https://github.com/mum4k/termdash/commit/e9b4ba36135dfb17dc27d61ab6ee21b8e3bd72b6

Let me know what you think.

keithknott26: Thanks, I've included the changes you've suggested in the latest commit.

widgets/indicator/options.go
keithknott26: Good point, I've spit the options up into Color and LabelColor for readability.

widgets/indicator/indicator.go
keithknott26: Very good suggestions, I've included these changes in the latest commit.

mum4k:
Hi @keithknott26, I am afraid I have dropped the ball on this PR and as a result I have also lost all context on it in the meantime.

Are you still interested in getting this in, should we revive the review?

mum4k:
Closing for now, but please feel free to reopen if you are still interested in getting this in.

PR #211, opened 2019-05-21
build explicitly disabling CGO

fenollp:


mum4k:
Thanks @fenollp for the contribution.

Would you mind sharing some background and / or motivation for the change?

fenollp:
Sure. It wasn't clear at a glance whether I could use this dependency while continuing to not use CGO (I'm trying to only have pure Go dependencies, mainly for easy crosscompilation).
So I tried to build explicitly disabling CGO and it works :) I thought this could be useful to others too.

mum4k:
That is a good idea @fenollp.

Thank you for the explanation.

Issue #100, opened 2019-02-01
Explore tcell as a termbox alternative


mum4k:
https://github.com/gdamore/tcell

We should add a new API wrapper like the one for termbox:
https://github.com/mum4k/termdash/tree/master/terminal/termbox

And document how users can choose termbox or tcell when initialising.

mum4k:
What is left is to solve the problem of color differences between termbox and tcell. We could approach this by adding a new color mode that will translate colors the same way termbox does. More details in the comment history of #224.

@kvnxiao, thank you for your help with #224, please let me know if you are also planning to work on the color differences, or if you would prefer I help with that.

kvnxiao:
I assume the colour differences would be a simple translation mapping between the specific colours that termbox uses. I will probably have time to update the wiki, but if it is alright with you I am hoping to ask you to help with the colour differences (I'm in my last semester of undergrad and things are ramping up for me).

mum4k:
Sounds good @kvnxiao, I will look at the color differences. Thank you very much for you help thus far and best of luck with your studies.

kvnxiao:
I've added a new wiki page here https://github.com/mum4k/termdash/wiki/Tcell-API but it's currently unlinked and not in the sidebar as the changes have not been released yet.

PR #224, opened 2020-02-29
Add Tcell implementation for terminal backend

### LINKED ISSUE #223: Does not render on Windows Terminal (preview)
Currently using https://github.com/microsoft/terminal from the window's app store (which is still in preview since it has not officially released yet) on my Windows machine.

I'm trying to make a cross-platform terminal application but it doesn't seem to work at all through Windows Terminal, regardless of the underlying shell I use (e.g. neither cmd nor powershell work when launched through Windows Terminal).

I am aware that termbox-go does not work through mintty or cygwin, **but I can confirm that termbox-go and termui _do_ work through the new Windows Terminal application**.

Here is an example of a termui application running in windows terminal:
image

However when I try to build the [TextInputDemo](https://github.com/mum4k/termdash/blob/master/widgets/textinput/textinputdemo/textinputdemo.go) as an executable and run this through Windows Terminal, it does not show up at all.

image

It simply shows a flashing cursor at the top left.

kvnxiao:
Adds the option to use https://github.com/gdamore/tcell as the terminal screen. This PR also addresses #223, being that tcell enables proper rendering on Windows devices (including Windows Terminal).

Works on #100.
Fixes #223.


kvnxiao:
@googlebot I signed it!

terminal/tcell/cell_options.go
```diff
@@ -0,0 +1,62 @@
+// Copyright 2020 Google Inc.
```
mum4k: What are your thoughts on updating the wiki as discussed in #223? Will you have the time and will to do that or would you prefer if I do?

terminal/tcell/cell_options.go
```diff
@@ -0,0 +1,62 @@
...
+//
```
mum4k: Can you please also update the [CHANGELOG.md](https://github.com/mum4k/termdash/blob/master/CHANGELOG.md) file in this PR? We can list this as "Added" functionality in the Unreleased section.

terminal/tcell/event.go
```diff
@@ -0,0 +1,192 @@
...
+	if tcellKey == tcell.KeyRune {
+		ch := event.Rune()
+		return &terminalapi.Keyboard{
+			Key: keyboard.Key(ch),
+		}
+	}
+
+	k, ok := tcellToTd[tcellKey]
+	if !ok {
+		return terminalapi.NewErrorf("unknown keyboard key '%v' in a keyboard event", tcellKey)
```
mum4k: (optional) Would it be better to print event.Name() instead? Not sure which one would be more readable.

https://github.com/gdamore/tcell/blob/de7e78efa4a71b3f36c7154989c529dbdf9ae623/key.go#L206

terminal/tcell/event.go
```diff
@@ -0,0 +1,192 @@
...
+	switch tcellBtn = event.Buttons(); tcellBtn {
+	case tcell.ButtonNone:
+		button = mouse.ButtonRelease
+	case tcell.Button1:
+		button = mouse.ButtonLeft
+	case tcell.Button2:
+		button = mouse.ButtonRight
+	case tcell.Button3:
+		button = mouse.ButtonMiddle
+	default:
```
mum4k: While this is according to what we discussed, the way it is implemented now the conversion layer would still forward an empty terminalapi.Mouse{} event with the default button and some x and y coordinates. This may confuse the upstream. A better approach might be to completely drop the event.

We might be able to accomplish that by modifying this function's signature so that it returns (terminalapi.Event, bool) and documenting that the bool is true iff we received an event that is valid for termdash. The caller then can check the bool and only forward the event if it is true.

terminal/tcell/event.go
```diff
@@ -0,0 +1,192 @@
...
+		// Do nothing, since tcell allows multiple buttons to be pressed at the same time
```
mum4k: Let's also add test coverage for this case once we address the comment above.

terminal/tcell/tcell.go
```diff
@@ -0,0 +1,205 @@
...
+		Y: h,
+	}
+}
+
+// Clear implements terminalapi.Terminal.Clear.
+func (t *Terminal) Clear(opts ...cell.Option) error {
+	o := cell.NewOptions(opts...)
+	st := cellOptsToStyle(o, t.colorMode)
+	w, h := t.screen.Size()
+	for row := 0; row < h; row++ {
```
mum4k: Can we use the Fill() method on the screen object instead or do we need to iterate ourselves?

https://github.com/gdamore/tcell/blob/b03a4607a44a5d8b4d4a0b9d2fde3ced0c28e78c/screen.go#L33

kvnxiao: :+1: Will look into it, originally used the code from tcell's termbox compat file which manually iterated and set the content to ' '.

terminal/tcell/event.go
kvnxiao: What would we do for callers of `convMouse()` then?

The only and immediate caller of `convMouse()` is `func toTermdashEvents(...) []terminalapi.Event` which returns a slice of `terminalapi.Event`s

If we change the signature of `convMouse()` to return -> `(terminalapi.Event, bool)`, how do we handle it from the toTermdashEvents function?

kvnxiao: For example, suppose that in https://github.com/mum4k/termdash/pull/224/files#diff-740d1ea120817d115c83619bb666a475R179 the code is changed to

```go
	case *tcell.EventMouse:
		mouseEvent, termdashOk := convMouse(event)
		...
		return []terminalapi.Event{mouseEvent}
```
and termdashOk is a boolean where convMouse returns true for all valid termdash event cases (and false for that default: case, e.g. where multiple mouse buttons are pressed at the same time)

What should be done if termdashOk == false? Since there needs to be a default case to match with the signature of `toTermdashEvents` returning `[]terminalapi.Event`. We can't have a `terminalapi.NewErrorf` event since that would crash the application.

terminal/tcell/tcell.go
kvnxiao: Switched to Fill() and it works fine.

terminal/tcell/cell_options.go
kvnxiao: Sure I can probably update the wiki, at least after this PR is merged

terminal/tcell/event.go
mum4k: There is no direct connection between `toTermdashEvents()` and the main processing loop of Termdash. In fact they run in separate goroutines and there is an unbound queue in between them. The main processing loop of termdash receives events by periodically calling the `Event()` method on the terminal API:

https://github.com/mum4k/termdash/blob/68bf0566e6ec8effaf6e83400ef0046347a5f5f1/termdash.go#L309-L310

Because it is written this way, the main processing loop won't mind if events aren't available on every call, it will simply retry later (it is a bit more complicated, the main processing loop actually gets notified when the next event is available, but that isn't relevant here). Therefore all we need to do is simply avoid push the unsupported event onto the unbound queue.

Btw it just occurred to me that we may not need any special treatment or a boolean in the function signature, we can simply return a nil from `convMouse()` directly and look for that inside `toTermdashEvents()`.

And when `convMouse()` returns nil, `toTermdashEvents()` can also return nil (which is equivalent to an empty slice in Go, i.e. a func that returns the []foo type can return nil). The result of that should be that `pollEvents()` will get no events and thus push no events onto the unbound queue.

Please double check this logic, I may be missing something, but it does seem to check out.

terminal/tcell/cell_options.go
mum4k: Doing this after we merge it in sounds good to me, thank you for offering and I am happy to help you with that, just let me know.

kvnxiao:
The mouse button change has been implemented. All that's left is to change the CHANGELOG, and code coverage. But I'm not exactly sure what code coverage is complaining about (I've already added tests for the private helper functions for `cell_options.go`).

mum4k:
We don't need to worry about the code coverage warning. Some part of the added code contains tcell functions that cannot be easily tested, since tcell doesn't seem to have convenient fakes. I think at this point you have done all that can be done for code coverage on this PR, so we can just ignore that warning when merging.

I will review your latest commits shortly and we can move forward.

terminal/tcell/event_test.go
```diff
@@ -112,26 +112,32 @@ func TestToTermdashEvents(t *testing.T) {
...
 		t.Run(fmt.Sprintf("key:%v want:%v", tc.btnMask, tc.want), func(t *testing.T) {
 
 			evs := toTermdashEvents(tcell.NewEventMouse(0, 0, tc.btnMask, tcell.ModNone))
-			if got, want := len(evs), 1; got != want {
+			if got, want := len(evs), len(tc.want); got != want {
 				t.Fatalf("toTermdashEvents => got %d events, want %d", got, want)
 			}
 
+			// Events that may exist for the terminal implementation but are not valid for termdash will be nil
+			if len(tc.want) == 0 {
```
mum4k: (nit, optional) While the table test structure suggests that test cases here can be written as expecting multiple mouse events (the want value is now a slice), the test logic verifies only the first event in that list (tc.want[0] nelow). That is fine for our usage, but might become confusing in the future.

Can we add a check. so that if future authors specify more than one event in tc.want, they get a readable error message instead of the other events being ignored? We could change this if statement as follows:

```go
switch count := len(tc.want); count {
case 0:
  // Events that may exist for the terminal implementation but are not valid for termdash will be nil
  return
case 1:
  // Proceed with the test.
default:
  t.Fatalf("toTermdashEvents test case specified %d expected events, only one is supported", count)
}
```

terminal/tcell/tcell.go
```diff
@@ -0,0 +1,200 @@
+// Copyright 2020 Google Inc.
```
mum4k: Probably missed from the previous round, posting as a standalone comment so we an discuss:

I am wondering if we should also update the termdashdemo. We could add a command line argument that would allow running it optionally with tcell for an easy comparison. What do you think? This would also allow me to easily test and verify functionality of the tcell layer.

kvnxiao: Would we only want to update termdashdemo? What about the other widget demos?

kvnxiao: I am 👍  for adding a flag to switch the implementation.

kvnxiao:
I just realized that https://github.com/mum4k/termdash/blob/master/terminal/terminalapi/terminalapi.go does not have a `Close()` function defined in the interface. Should I add that in as a fix? Because the comments for both tcell and termbox implementations of their Close() function say that it implements `terminalapi.Terminal.Close` which is currently incorrect.

terminal/tcell/tcell.go
mum4k: That is a good question, I think eventually we will want to add this to all the demos and since tcell does seem better and assuming no negative performance implications - we might as well make it the default.

However we don't need to do it in this PR, it would just extend its scope. I would still go for termdashdemo though, so that we can play around with tcell a bit and maybe also profile the two and see differences in CPU and memory usage.

termdashdemo/termdashdemo.go
```diff
@@ -18,7 +18,9 @@ package main
```
mum4k: Thank you for adding this, this allowed me to test both and get a comparison. It looks like tcell might be running in a different color mode than termbox (or the colors are misaligned). Please look at the following two screenshots. I am on MacOs so it might be specific to this platform. Are you also seeing this difference? How can we address it?

Termbox:
termbox

Tcell:
tcell

kvnxiao:
Hmm, I think this issue is due to how tcell defines its colours. If I recall correctly and referring to https://jonasjacek.github.io/colors/ , the "8 system colours" which we define in `color.go` are the following:

```go
// The supported terminal colors.
const (
	ColorDefault Color = iota

	// 8 "system" colors.
	ColorBlack
	ColorRed
	ColorGreen
	ColorYellow
	ColorBlue
	ColorMagenta
	ColorCyan
	ColorWhite
)
```

When using any of the specified colours above and in 256 color mode... ignoring ColorDefault, since we subtract 1 from the specified termdash cell.Color to convert to tcell color:

```go
// cellColor converts termdash cell color to the tcell format.
func cellColor(c cell.Color) tcell.Color {
	return tcell.Color(c&0x1ff) - 1
}
```

The "8 system colours" simply map to `xterm-colors` 0-7 directly, which are the "dark variants" of the main 8 system colours, whereas termbox uses 8-15 hence the brighter variants of the main 8 system colours.

kvnxiao:
I'm not exactly sure what's the best way to go about fixing this... Since when we pass in a `cell.Color` in 256 color mode, tcell supports all the 256 colours defined by xterm-256, and it's very easy to convert directly from `cell.Color` to `tcell.Color` with just an off-by-one subtraction.

termdashdemo/termdashdemo.go
kvnxiao: Do you have anything in mind on how we could tackle the colour differences between termbox and tcell?

mum4k:
Thank you very much for explaining the behavior. I agree with you that we should not prevent the user from selecting all the colors (0-7, 8-15, ...). Now that we understand why this happens, I think we can proceed with this PR as is.

We could file an issue to track the color difference and solve it by defining a new ColorMode, something like "Termbox Compatible" mode. When this mode is selected and tcell is used as the terminal driver, we would translate between 0-7 and 8-15 in a way that would match termbox. Do you think this solution would work?

As mentioned we can and probably should address this outside of this PR to keep it focused. Please let me know what you think.

kvnxiao:
That sounds good to me! Let's defer the colour differences for a future PR then, and document the gotchas in the wiki.

mum4k:
Turns out this was a red herring. The color differences were due to my local terminal profile. Both termbox and tcell produce exactly the same output on a vanilla terminal.

PR #225, opened 2020-03-07
Release v0.11.0.

### LINKED ISSUE #100: Explore tcell as a termbox alternative

### LINKED ISSUE #223: Does not render on Windows Terminal (preview)

mum4k:
- New [`tcell`](https://github.com/gdamore/tcell) based terminal implementation.

Fixes #100.
Fixes #223.


mum4k:
The author of #224 signed CLA in that PR with the intention of merging this into master.

Issue #245, opened 2020-09-30
Windows console not drawing scroll arrows


d-ank:
https://github.com/mum4k/termdash/blob/fe3e973531917978dd6d412520528bf10e092087/widgets/text/text.go#L128
https://github.com/mum4k/termdash/blob/fe3e973531917978dd6d412520528bf10e092087/widgets/text/text.go#L147
[These characters are unable to be drawn in the windows console, or at least in the current rendition.](https://i.imgur.com/FwhQTdM.png)
maybe give an option to the text widget to swap the scroll up and down arrows out for ▲ and ▼ (or ↑ and ↓)?

mum4k:
Hi @d-ank, thank you for the suggestion. I agree that this is a good solution and should be fairly trivial to implement.

Is this something you would be willing to contribute?

Amirhossein2000:
Hi
```Go
cells, err := cvs.SetCell(cur, '⇧') --> cells, err := cvs.SetCell(cur, ' ▲')
```
Is this as easy as this or I'm wrong?

d-ank:
Sorry for the delayed response, I will put in a pr in when I am at my desktop.

mum4k:
@Amirhossein2000 that is the essence of the code change, yes. However if done like this, it would change the default behavior for all the users. What might be better is for us to add an option that would allow the user to specify the character to use in place of the arrows. One example of implementing such option is in the Gauge widget that allows setting a custom Gauge character:

https://github.com/mum4k/termdash/blob/fe3e973531917978dd6d412520528bf10e092087/widgets/gauge/options.go#L83-L89

@d-ank no pressure, take your time and thank you for looking into this.

d-ank:
https://github.com/mum4k/termdash/pull/248 Here is the accompanying fix and addition.

mum4k:
Released in #266.

PR #248, opened 2020-10-13
Ability to swap scrolling runes 

d-ank:
#245 fixes this issue with windows consoles by adding option to switch the scroll runes on a text widget.
The option `text.ScrollRunes` accepts two rune parameters, up and down, the default runes are the same `⇧` and `⇩`.

An example of this which will draw correct scroll runes on windows
`text.New(text.RollContent(), text.WrapAtWords(), text.ScrollRunes('▲', '▼'))`

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,6 +209,7 @@ Termdash uses [this branching model](https://nvie.com/posts/a-successful-git-bra
 - [datadash](https://github.com/keithknott26/datadash): Visualize streaming or tabular data inside the terminal.
 - [grafterm](https://github.com/slok/grafterm): Metrics dashboards visualization on the terminal.
 - [perfstat](https://github.com/flaviostutz/perfstat): Analyze and show tips about possible bottlenecks in Linux systems.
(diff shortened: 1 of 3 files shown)


d-ank:
@googlebot I signed it!

mum4k (changes requested):
Looks spot on, thank you for the contribution.

Would you be also willing to add test coverage for the new options? It should be as simple as copying and slightly modifying some of the existing test cases here:

https://github.com/mum4k/termdash/blob/master/widgets/text/text_test.go

You can try searching for "scroll marker" to find test cases that exercise behaviors resulting in the scroll arrows being drawn. We can add two test cases, one that draws the bottom marker and one that draws the top one.

d-ank:
Sorry I won't be able to add the test coverage till the weekend.

mum4k:
No rush @d-ank, please take your time.

mum4k:
Hi @d-ank, just checking in, are you still interested in adding the tests. Or would you prefer if we merge it in as is and I add the test coverage separately?

mum4k:
Thanks again @d-ank for the contribution, I will merge this in and add the tests.

PR #251, opened 2020-11-12
add font modifier cell options: bold, italic, underline.

dyc3:
This PR adds cell options for the font modifiers﻿ bold, italic, and underline.

`tcell` was upgraded 1.4.0.

Strikethrough cell option was added as a no-op until tcell is updated to v2.0.0. I figured it wasn't a good idea to upgrade tcell by a whole major version without testing.

Closes #251

### DIFF
--- a/cell/cell.go
+++ b/cell/cell.go
@@ -23,8 +23,12 @@ type Option interface {
 
 // Options stores the provided options.
 type Options struct {
-	FgColor Color
-	BgColor Color
+	FgColor       Color
+	BgColor       Color
+	Bold          bool
+	Italic        bool
+	Underline     bool
+	Strikethrough bool
 }
 
 // Set allows existing options to be passed as an option.
@@ -60,5 +64,33 @@ func FgColor(color Color) Option {
 func BgColor(color Color) Option {
 	return option(func(co *Options) {
 		co.BgColor = color
+	})
+}
+
+// Bold makes cell's text bold.
+func Bold() Option {
+	return option(func(co *Options) {
+		co.Bold = true
+	})
+}
+
+// Italic makes cell's text italic. Only works when using the tcell backend.
+func Italic() Option {
+	return option(func(co *Options) {
+		co.Italic = true
+	})
+}
+
+// Underline makes cell's text underlined.
+func Underline() Option {
+	return option(func(co *Options) {
+		co.Underline = true
+	})
+}
+
+// Strikethrough strikes through the cell's text. Currently a no-op until tcell is updated to >= v2.0.0
+func Strikethrough() Option {
+	return option(func(co *Options) {
+		co.Strikethrough = true
 	})
 }
--- a/terminal/termbox/cell_options.go
+++ b/terminal/termbox/cell_options.go
@@ -17,6 +17,8 @@ package termbox
 // cell_options.go converts termdash cell options to the termbox format.
 
 import (
+	"errors"
+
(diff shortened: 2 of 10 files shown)


terminal/termbox/cell_options.go
```diff
@@ -28,7 +28,16 @@ func cellColor(c cell.Color) tbx.Attribute {
 
 // cellOptsToFg converts the cell options to the termbox foreground attribute.
 func cellOptsToFg(opts *cell.Options) tbx.Attribute {
-	return cellColor(opts.FgColor)
+	a := cellColor(opts.FgColor)
+	if opts.Bold {
+		a |= tbx.AttrBold
+	}
+	// FIXME: Termbox doesn't have an italics attribute
```
mum4k, line 35: How would you feel about returning an error here when the user tried to set italics or strikethrough on termbox? Giving them complete information in an error message might be more user friendly than silently ignoring the requested options.

cell/cell.go
```diff
@@ -23,8 +23,12 @@ type Option interface {
 
 // Options stores the provided options.
 type Options struct {
-	FgColor Color
-	BgColor Color
+	FgColor       Color
```
mum4k, line 26: Would you be also willing to add test coverage for the modified modules? There are pre-existing unit test that should provide some guidance.

mum4k (changes requested):
Looks great, thank you for the contribution!

Can you please sign the CLA? In addition to that, can you please respond indicating that you are giving me permission to merge this into the master branch?

dyc3:
I thought I did sign the cla... I'll try it again

dyc3:
@googlebot I signed it!

cell/cell.go
dyc3: Sure thing!

terminal/termbox/cell_options.go
dyc3: Yeah, that sounds a lot better.

dyc3:
> can you please respond indicating that you are giving me permission to merge this into the master branch?

I give you permission to merge this into the master branch.

mum4k:
Thank you for the permission. We will merge this into the devel branch first and it will go into the next release together with a few minor changes.

mum4k:
@dyc3 please let me know if you are still working on this, or if it is ready for merging. There is still some missing coverage on some of the added code, but I will be happy to add those tests as a follow-up if you prefer that.

dyc3:
I'm currently done working on this. I looked at the missing coverage, and I thought the behavior that wasn't being covered was pretty trivial and wouldn't benefit too much from tests.

Although looking at it again, should the error paths be covered in tests?

mum4k:
I would cover the error paths that we can cover easily. But there certainly are some not worth the work, because they depend on the termbox object being instantiated and we don't have a pre-established way of doing that so far.

I think this is good, I propose to merge it as it is and I can try to iterate on the test coverage a bit more before the merge to master.

Thank you very much for improving Termdash. I will aim towards upgrading tcell to 2.x next and then we can add support for the additional font modifiers. Please feel free to let me know if you would like to help with either of the two.

dyc3:
Agree, lets merge it as-is. 

I'd like to help out with both. Feel free to @ me for whatever needs to get done.

mum4k:
Sounds good and will do.

mum4k (approved):

PR #260, opened 2020-11-14
Upgrade tcell to v2.x

### LINKED ISSUE #254: Update tcell to the latest 2.x version
Changelog: https://github.com/gdamore/tcell/blob/master/CHANGESv2.adoc

And while we are at it, we can also consider making `tcell` the default in all code examples.

dyc3:
A draft PR to close #254 

I'll push stuff here as I get stuff working.


CHANGELOG.md
```diff
@@ -13,7 +13,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- `tcell` dependency was upgraded 1.4.0.
+- `tcell` dependency was upgraded v2.0.0.
```
mum4k, line 16: (nit) I made a typo here, missing "to" in ... to v2.0.0 in case you don't mind fixing it.

terminal/tcell/event.go
```diff
@@ -17,7 +17,7 @@ package tcell
 import (
 	"image"
 
-	"github.com/gdamore/tcell"
+	"github.com/gdamore/tcell/v2"
```
mum4k, line 20: Would you mind specifically naming this as "tcell"?

```go
tcell "github.com/gdamore/tcell/v2"
```

I use goimports to organize the imports and without the rename goimports just removes the import altogether, since it doesn't see "v2" being used in the code. Maybe also be a bug in the goimports version I use? Do you know?

In case we change this, we should do so in all the locations.

mum4k (changes requested):
Looks good, just two asks and one of them is on my own code ... sorry.

terminal/tcell/event.go
dyc3: I'm using whatever the defaults are in vscode and I don't have that issue.

mum4k: That suggests this will be problem with my setup. Thanks for renaming it for now.

mum4k (approved):

PR #266, opened 2020-11-18
Releasing Termdash v0.13.

mum4k:
### Added

- the `Text` widget now allows user to specify custom scroll marker runes.

### Changed

- terminal cells now support font modifier options (bold, italic,
  underline, strike through).
- `tcell` dependency was upgraded to v2.0.0.
- upgraded versions of all other dependencies.
- aligned the definition of the first 16 colors with the definition used by
  Xterm and `tcell`. Defined two non-standard colors `ColorMagenta` and
  `ColorCyan` to make this change backward compatible for users that use
  `termbox-go`.
- made `tcell` terminal implementation the default in examples, demos and
  documentation.

### Fixed

- coveralls again triggers and reports on PRs.
- addressed some lint issues.
- improved test coverage in some modules.
- changed the Blue color in demos to a more visible shade.
- fixed a bug where segment display text in `termdashdemo` appeared to be
  jumping.


mum4k:
Individual contributors specifically intended for their code to go into the master branch and confirmed so by signing CLAs on the individual PRs.

PR #272, opened 2020-11-25
Give metadata about events to widgets.

mum4k:
Widgets now receive an additional argument together with keyboard and mouse events that provides metadata about the event. Initially it indicates if the widget's container was focused at the time of the event.


mum4k:
After further consideration, this isn't really necessary. The use case I had in mind can be achieved without this metadata and I don't know of another. The lack of use case doesn't justify the breaking API change which should be avoided until really necessary.

PR #278, opened 2020-12-05
Implementing the HeatMap widget

### LINKED ISSUE #265: Widget: heat map
Hello, I implemented a heat map widget with termdash in [SkyWalking-CLI](https://github.com/apache/skywalking-cli), like this:
heatmap

I want to contribute it back recently, but I need some time to polish code to make it more general. I think the format of input data should be `[['xlabel', 'ylabel', value],...]`, how about you ? 

If you have any advice about the widget, please let me know :)

fgksgf:
- [x] Define the API
- [x] Implement basic features
- [x] Implement `playHeatMap` for demo
- [ ] Write tests [WIP]
- [ ] Polish code and comments

Close #265


widgets/heatmap/options.go
```diff
@@ -29,21 +28,25 @@ type Option interface {
...
+	hideXLabels    bool
+	hideYLabels    bool
 	xLabelCellOpts []cell.Option
 	yLabelCellOpts []cell.Option
 }
 
 // validate validates the provided options.
 func (o *options) validate() error {
-	return errors.New("not implemented")
+	//if got, min := o.cellWidth, DefaultCellWidth; got < min {
```
mum4k, line 40: This block might be left commented out by accident?

Applies to lines 49 and 65 below too:
```go
opt := &options{
    //cellWidth: DefaultCellWidth,
}

//const DefaultCellWidth = 3
```

widgets/heatmap/heatmapdemo/heatmapdemo.go
```diff
@@ -24,21 +24,86 @@ import (
...
 		panic(err)
 	}
 
-	// TODO: set heatmap's data
+	if err := hp.Values(getData()); err != nil {
+		panic(err)
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	//go playHeatMap(ctx, hp, 1*time.Second)
```
mum4k, line 106: You might already be aware of this, the current implementation, the current implementation seems to have a bug. When we uncomment this line, we get the following failure after a couple of updates:

```go
panic: container.Draw => error: unable to draw widget *heatmap.HeatMap: unable to set cell runes in area (74,0)-(86,1), it must fit inside the available cell area is (0,0)-(78,23)

goroutine 1 [running]:
main.main()
	/Users/mum4k/go/src/github.com/mum4k/termdash/widgets/heatmap/heatmapdemo/heatmapdemo.go:127 +0x88c
exit status 2
```

mum4k, line 106: Additionally when I downsize (resize) the terminal window with the heatmap demo, the application crashes and we get:

```go
panic: container.Draw => error: unable to draw widget *heatmap.HeatMap: unable to set cell runes in area (66,0)-(78,1), it must fit inside the available cell area is (0,0)-(75,22)
```

I perform these tests, because most of the widgets that I have implemented had a problem with this initially. These bugs usually meant that I didn't include the necessary size adjustment or error handling in the Draw method, or that the widget didn't specify the correct options back to the infrastructure. Please feel free to reach our if you will have trouble finding the issue, we can debug it together.

mum4k (changes requested):
This looks great! Not performing a code review yet, because we have a couple of usability bugs / crashes (see below). Additionally I would suggest we update the demo to display some form of dynamic data so we can further verify the stability of the widget's implementation before we start doing a code level review.

Please feel free to reach our at any point if you would like my help or opinion on any of these. Your contribution is greatly appreciated.

widgets/heatmap/heatmapdemo/heatmapdemo.go
fgksgf: Resolved

fgksgf: Resolved, but I found a new bug. The first column of the canvas can't be displayed:
- should be `PRESS` instead of `RESS`
- only show one digit of the Y labels 
image

widgets/heatmap/heatmap.go
```diff
@@ -89,42 +144,161 @@ func (hp *HeatMap) ClearYLabels() {
...
+	}
+
+	return xd, yd, nil
 }
 
 // Draw draws cells, X labels and Y labels as HeatMap.
 // Implements widgetapi.Widget.Draw.
 func (hp *HeatMap) Draw(cvs *canvas.Canvas, meta *widgetapi.Meta) error {
-	return errors.New("not implemented")
+	hp.mu.Lock()
```
mum4k, line 196: Thank you for continuing to work on this. On top of the bug that you identified, we also seem to have one when the widget's window is resized down to just 7 characters of width:

```
panic: runtime error: integer divide by zero

goroutine 1 [running]:
github.com/mum4k/termdash/widgets/heatmap.(*HeatMap).cellWidthAdaptive(0xc000166120, 0xc0002049c0)
	/Users/mum4k/go/src/github.com/mum4k/termdash/widgets/heatmap/heatmap.go:275 +0xb8
github.com/mum4k/termdash/widgets/heatmap.(*HeatMap).axesDetails(0xc000166120, 0xc0002049c0, 0x0, 0x0, 0x3, 0x19)
	/Users/mum4k/go/src/github.com/mum4k/termdash/widgets/heatmap/heatmap.go:178 +0x39
github.com/mum4k/termdash/widgets/heatmap.(*HeatMap).Draw(0xc000166120, 0xc0002049c0, 0xc0004fc810, 0x0, 0x0)
	/Users/mum4k/go/src/github.com/mum4k/termdash/widgets/heatmap/heatmap.go:210 +0x14c
github.com/mum4k/termdash/container.drawWidget(0xc00013e700, 0x0, 0x0)
	/Users/mum4k/go/src/github.com/mum4k/termdash/container/draw.go:138 +0x2bb
github.com/mum4k/termdash/container.drawCont(0xc00013e700, 0x1, 0x1)
	/Users/mum4k/go/src/github.com/mum4k/termdash/container/draw.go:171 +0x147
github.com/mum4k/termdash/container.drawTree.func1(0xc00013e700, 0x7, 0x1b)
	/Users/mum4k/go/src/github.com/mum4k/termdash/container/draw.go:63 +0x85
github.com/mum4k/termdash/container.preOrder(0xc00013e700, 0xc000265bd8, 0x119d550)
	/Users/mum4k/go/src/github.com/mum4k/termdash/container/traversal.go:43 +0x4f
github.com/mum4k/termdash/container.drawTree(0xc00013e700, 0x1b, 0x0)
	/Users/mum4k/go/src/github.com/mum4k/termdash/container/draw.go:43 +0x125
github.com/mum4k/termdash/container.(*Container).Draw(0xc00013e700, 0x0, 0x0)
	/Users/mum4k/go/src/github.com/mum4k/termdash/container/container.go:233 +0x11c
github.com/mum4k/termdash.(*termdash).redraw(0xc000134420, 0x0, 0x0)
	/Users/mum4k/go/src/github.com/mum4k/termdash/termdash.go:274 +0x3d
github.com/mum4k/termdash.(*termdash).periodicRedraw(0xc000134420, 0x0, 0x0)
	/Users/mum4k/go/src/github.com/mum4k/termdash/termdash.go:301 +0x71
github.com/mum4k/termdash.(*termdash).start(0xc000134420, 0x11c4f40, 0xc000137100, 0x0, 0x0)
	/Users/mum4k/go/src/github.com/mum4k/termdash/termdash.go:344 +0x216
```

widgets/heatmap/heatmapdemo/heatmapdemo.go
mum4k: This one looks interesting. I tried playing with the demo, but was unable to reproduce it. How can we get it into this state?

Also as mentioned before - please feel free to let me know if you would like my help debugging any of these issues. I will gladly cooperate here.

fgksgf: I run all demos like this. But when I use `termbox` instead of `tcell` in `heatmapdemo.go`, it becomes normal.

mum4k: Good catch, I managed to reproduce this in other demo and filed #281. Feel free to develop this with `termbox` to avoid the issue. I will work on tracking it down.

mum4k: Fyi #283 fixes the issue.

fgksgf:
hello, I started writing test code, and I think it's ok to review the code :)

mum4k:
Great news @fgksgf, I will start reading through it and point out if I have any suggestions.

Looks like this no longer crashes as it did before with `tcell` now that #281 has been resolved. We do however seem to have a bug that might be in the code of `heatmap`. Reproduce by running the demo and decrease the window size vertically to less than 2 cells. I am using iTerm on MacOS which allows such small size. This might be just incorrectly set limits? This is the traceback:

```
mum4k@elephant:heatmapdemo (fgksgf-heatmap)$ cat /tmp/log
panic:

goroutine 37 [running]:
main.playHeatMap(0x11c58c0, 0xc0000b0080, 0xc0000d0000, 0x3b9aca00)
        /Users/mum4k/go/src/github.com/mum4k/termdash/widgets/heatmap/heatmapdemo/heatmapdemo.go:53 +0x437
created by main.main
        /Users/mum4k/go/src/github.com/mum4k/termdash/widgets/heatmap/heatmapdemo/heatmapdemo.go:75 +0x196
exit status 2
```

Could you take a look at this? The good news is that this seems to be the last crash related to terminal resize which in my experience are the hardest to get right.

widgets/heatmap/options.go
```diff
@@ -15,8 +15,9 @@
 package heatmap
```
mum4k, line 15: (nit) We may want to update the licence headers to say 2021 since it looks like it will land in that year.

widgets/heatmap/internal/axes/label.go
```diff
@@ -36,21 +38,85 @@ type Label struct {
...
+
+	return &Label{
+		Text: label,
+		Pos:  pos,
+	}, nil
 }
 
 // xLabels returns labels that should be placed under the cells.
 // Labels are returned with X coordinates in ascending order.
 // X coordinates grow right.
```
mum4k, line 77: Can we say something about the `yEnd` and `graphWidth` arguments to help any future readers?

widgets/heatmap/internal/axes/label.go
```diff
@@ -36,21 +38,85 @@ type Label struct {
...
+// rowLabel returns one Y label for the specified row.
 // The row is the Y coordinate of the row, Y coordinates grow down.
 func rowLabel(row int, label string, labelWidth int) (*Label, error) {
-	return nil, errors.New("not implemented")
+	// The area available for the Y label
+	ar := image.Rect(0, row, labelWidth, row+1)
+
+	pos, err := alignfor.Text(ar, label, align.HorizontalRight, align.VerticalMiddle)
+	if err != nil {
+		return nil, fmt.Errorf("unable to align the label value: %v", err)
```
mum4k, line 66: We have two error messages here that both say the exact same thing:

`unable to align the label value`

Can we add information about which label they are talking about to help disambiguate between them? E.g:

For here: `unable to align the Y label value: %v`
For the one below: `unable to align the X label value: %v`

widgets/heatmap/internal/axes/label.go
```diff
@@ -36,21 +38,85 @@ type Label struct {
...
 // Labels are returned with X coordinates in ascending order.
 // X coordinates grow right.
 func xLabels(yEnd image.Point, graphWidth int, labels []string, cellWidth int) ([]*Label, error) {
-	return nil, errors.New("not implemented")
+	var ret []*Label
+
+	// The length of the longest X labels.
+	ls := LongestString(labels)
+
+	if ls > cellWidth {
```
mum4k, line 84: It is fairly unusual for idiomatic Go code to have many levels of indentation on non-error flow:

https://gist.github.com/adamveld12/c0d9f0d5f0e1fba1e551#indent-error-flow

Since the two cases are very distinct, how would you feel about extracting the inner code (the for loops) into two distinct private helper functions? That way the `xLabels` functions ends up containing just the high level business logic, will be easier to read and contain just a single level of indentation.

Example:

```go
if ls > cellWidth {
  ret, err = alignSomeLabels()
} else {
  ret, err = alignAllLabels() 
}
```

widgets/heatmap/internal/axes/label.go
```diff
@@ -60,5 +126,13 @@ func xLabels(yEnd image.Point, graphWidth int, labels []string, cellWidth int) (
 // the X label belongs to the middle column of the three columns,
 // and the padded length is 3*3 (cellWidth multiplies the number of columns), which is 9.
 func paddedLabelLength(graphWidth, longest, cellWidth int) (l, index int) {
```
mum4k, line 128: Considering that the scope of the `l` variable here is the entire function and it is also a named return parameter, can we give it a bit more readable name? How about 'length` instead?

mum4k (changes requested):
Partial review, reviewed the following files:

- options.go
- label.go

Will review the rest in the next rounds. From what I have seen so far, this PR contains very clean and easy to read code, so I have a minimal amount of comments thus far. Thank you for making it so.

fgksgf:
> Could you take a look at this? The good news is that this seems to be the last crash related to terminal resize which in my experience are the hardest to get right.

Fixed.  I'm considering if we should rename `cell` in the code of  `heatmap` to `grid` ? Because the `cell` has been used and may cause ambiguity. What do you think ?

widgets/heatmap/heatmap.go
```diff
@@ -76,7 +88,50 @@ func New(opts ...Option) (*HeatMap, error) {
...
-	return errors.New("not implemented")
+	if len(values) == 0 {
+		return errors.New("the values cannot be empty")
+	}
+
+	hp.mu.Lock()
+	defer hp.mu.Unlock()
+
+	var (
+		xl []string
```
mum4k, line 99: Since these variables have larger scope (the full function), the we give them longer (more readable) names?

widgets/heatmap/heatmap.go
```diff
@@ -76,7 +88,50 @@ func New(opts ...Option) (*HeatMap, error) {
...
+	hp.yLabels = yl
+	hp.values = v
+	for _, opt := range opts {
+		opt.set(hp.opts)
+	}
+	hp.minValue, hp.maxValue = minMax(hp.values)
+	return nil
 }
 
 // ClearXLabels clear the X labels.
```
mum4k, line 137: Can we ensure that all functions that change or read the state of the HeatMap do so with the mutex acquired? 

The functions `ClearXLabels` and `ClearYLabels` don't seem to be doing so currently which might lead to unexpected race conditions. Applies to all such functions.

widgets/heatmap/heatmap.go
```diff
@@ -76,7 +88,50 @@ func New(opts ...Option) (*HeatMap, error) {
 // Each call to Values overwrites any previously provided values.
 // Provided options override values set when New() was called.
 func (hp *HeatMap) Values(xLabels []string, yLabels []string, values [][]float64, opts ...Option) error {
```
mum4k, line 90: Just occurred to me that the coordinates in values are in the order (y, x) while most coordinates in termdash are in order `(x, y)` as inspired by the `image` library: https://godoc.org/image#Point.

Is the `(y, x)` order intentional? Do you have preferences? Or is it because we are treating it as a matrix of `(rows, cols)`? Whichever we choose, can we make sure it is well documented and the variable names correspond to the chosen order? I.e. if it is a matrix, then we can use the words `rows` and `cols` when referring to coordinates instead of generic counter names like `i` and `j`.

Related to another comment I made below.

widgets/heatmap/heatmap.go
```diff
@@ -89,42 +144,161 @@ func (hp *HeatMap) ClearYLabels() {
...
+func (hp *HeatMap) ValueCapacity() (rows, cols int) {
+	hp.mu.Lock()
+	defer hp.mu.Unlock()
+
+	if hp.lastWidth == 0 || hp.lastHeight == 0 {
+		return 0, 0
+	}
+
+	rows = hp.lastHeight - 1
+	var cw int
```
mum4k, line 165: The remaining section of this function could use some unwrapping to make it easier to read. We could start by giving `cw` a better name. Next we could unwrap the `cols` calculation into multiple statements that will either become self-explanatory, or we can add inline comments.

As it stands, I am having a hard time understanding the calculation. Can you help out?

widgets/heatmap/heatmap.go
```diff
@@ -89,42 +144,161 @@ func (hp *HeatMap) ClearYLabels() {
...
+
+// minCellWidth is the minimum width of each cell in the heat map.
+const minCellWidth = 3
+
+// cellWidthAdaptive determines the width of a single cell (grid) based the canvas.
+func (hp *HeatMap) cellWidthAdaptive(cvs *canvas.Canvas) {
+	rem := cvs.Area().Dx() - axes.LongestString(hp.yLabels) - axes.AxisWidth
+	var cw int
+	if len(hp.values) != 0 && len(hp.values[0]) != 0 {
+		cw = rem / len(hp.values[0])
```
mum4k, line 275: Do we guarantee (validate) that the shape of the values will always be a rectangle, i.e. if len(hp.values) == 2 then len(hp.values[0]) == len(hp.values[1]) ?

widgets/heatmap/heatmap.go
```diff
@@ -89,42 +144,161 @@ func (hp *HeatMap) ClearYLabels() {
...
+	}
+	if cw >= minCellWidth {
+		hp.opts.cellWidth = cw
```
mum4k, line 278: It makes a minor difference in readability, but it might be worth it. Would you mind making this function return the calculated value? That way the call site makes it clear what the function does. As it stands its execution has an unknown side effect which is non-obvious until the reader looks here.

widgets/heatmap/heatmap.go
```diff
@@ -89,42 +144,161 @@ func (hp *HeatMap) ClearYLabels() {
...
+		}
+	}
+	return nil
+}
+
+// minCellWidth is the minimum width of each cell in the heat map.
+const minCellWidth = 3
+
+// cellWidthAdaptive determines the width of a single cell (grid) based the canvas.
+func (hp *HeatMap) cellWidthAdaptive(cvs *canvas.Canvas) {
```
mum4k, line 271: I am wondering if we could improve the name `cellWidthAdaptive`. How does `adaptCellWidth` sound?

widgets/heatmap/heatmap.go
```diff
@@ -89,42 +144,161 @@ func (hp *HeatMap) ClearYLabels() {
...
+			if err := draw.Text(cvs, l.Text, l.Pos, draw.TextCellOpts(hp.opts.xLabelCellOpts...)); err != nil {
+				return fmt.Errorf("failed to draw the X labels: %v", err)
+			}
+		}
+	}
+	return nil
+}
+
+// minCellWidth is the minimum width of each cell in the heat map.
+const minCellWidth = 3
```
mum4k, line 268: Curious - why can't we have a cell width of 1 or 2?

widgets/heatmap/heatmap.go
```diff
@@ -89,42 +144,161 @@ func (hp *HeatMap) ClearYLabels() {
...
+	}
+	if !needAr.In(cvs.Area()) {
+		return draw.ResizeNeeded(cvs)
+	}
+
+	xd, yd, err := hp.axesDetails(cvs)
+	if err != nil {
+		return err
+	}
+	err = hp.drawCells(cvs, yd)
```
mum4k, line 214: We should always try to minimize the scope of the `err` variable:

```go
if err := hp.drawCells(cvs, yd); err != nil { ... } // Notice the := instead of just =.
```

If we minimize the scope, we avoid subtle bugs where we can forget to check the value of err. As the code stands now one could add another call, forget to check the value and the compiler won;t complain, because `err` was already used above.

widgets/heatmap/heatmap.go
```diff
@@ -89,42 +144,161 @@ func (hp *HeatMap) ClearYLabels() {
...
+
+	return hp.drawLabels(cvs, xd, yd)
 }
 
 // drawCells draws m*n cells (rectangles) representing the stored values.
 // The height of each cell is 1 and the default width is 3.
-func (hp *HeatMap) drawCells(cvs *canvas.Canvas, xd *axes.XDetails, yd *axes.YDetails) error {
-	return errors.New("not implemented")
+func (hp *HeatMap) drawCells(cvs *canvas.Canvas, yd *axes.YDetails) error {
+	for i := 0; i < len(hp.values); i++ {
```
mum4k, line 225: Since these are essentially coordinates, can we use `x` and `y` instead of `i` and `j`?

widgets/heatmap/heatmap.go
```diff
@@ -89,42 +144,161 @@ func (hp *HeatMap) ClearYLabels() {
...
+			if err := cvs.SetAreaCells(rect, hp.opts.cellChar, cell.BgColor(color)); err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
 }
 
 // drawAxes draws X labels (under the cells) and Y Labels (on the left side of the cell).
```
mum4k, line 245: Sorry, missed this before, the comment here should say `drawLabels` instead of `drawAxes`.

widgets/heatmap/heatmap.go
```diff
@@ -149,5 +323,31 @@ func (hp *HeatMap) Options() widgetapi.Options {
...
+// initLabels return initial labels, like '0', '1', '2', ...
+func initLabels(l int) []string {
+	var ret []string
+	for i := 0; i < l; i++ {
+		ret = append(ret, fmt.Sprintf("%d", i))
+	}
+	return ret
+}
+
+// minMax returns the min and max values in given integer array.
```
mum4k, line 341: Can we potentially use one of the functions defined in the `numbers` package?

https://github.com/mum4k/termdash/blob/74d9de20cbfdc89dc6255bc5a27f3335d0a6fe99/private/numbers/numbers.go#L99-L103

If not, can we modify or contribute another function to that package instead? That way these files can focus on just code directly related to the HeatMap.

mum4k (changes requested):
Sorry about the delay here, I was deeply in vacation mode. I should be more responsive from now on. Glad to hear you find the comments useful, please don't hesitate to let me know if that changes.

You raise a good point about `cell` being ambiguous, however `grid` might not be the best name, since grid indicates a larger collection of cell like entities, while we are trying to present each cell as a single unit to the end user. How about `unit` or `section`?

Continuing the partial review, the list of reviewed files is now:

- options.go
- label.go
- axes.go
- heatmap.go
- heatmapdemo.go

Ready to start reviewing tests once we address the comments here.

fgksgf:
> Sorry about the delay here, I was deeply in vacation mode. I should be more responsive from now on. 

Never mind, the collaboration in open source projects is inherently asynchronous, isn't it ? :)
 
> You raise a good point about `cell` being ambiguous, however `grid` might not be the best name, since grid indicates a larger collection of cell like entities, while we are trying to present each cell as a single unit to the end user. How about `unit` or `section`?

Make sense to me, I prefer `unit`.

widgets/heatmap/internal/axes/axes.go
```diff
@@ -42,17 +41,25 @@ type YDetails struct {
...
-// The parameter ls is the longest string in yLabels.
-func RequiredWidth(ls string) int {
-	return runewidth.StringWidth(ls) + axisWidth
-}
-
 // NewYDetails retrieves details about the Y axis required
 // to draw it on a canvas of the provided area.
 func NewYDetails(labels []string) (*YDetails, error) {
-	return nil, errors.New("not implemented")
+	// See how the labels would look like on the entire maxWidth.
```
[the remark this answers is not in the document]
fgksgf: copy-paste issue, removed

widgets/heatmap/heatmap.go
fgksgf: Makes sense 👍

mum4k:
Indeed, oss projects are asynchronous, tank you for your understanding.

With that said, please ping me on this PR once it is ready for another round, so I know when it is my turn ;)

widgets/heatmap/heatmap.go
fgksgf: I will add a check in `func Values()` to guarantee that.

widgets/heatmap/heatmap.go
fgksgf: Good catch. I use `(y, x)` because we treat values as a matrix of `(rows, cols)`.

widgets/heatmap/heatmap.go
fgksgf: Didn't notice the `numbers` package, contributed a new function `MinMaxOfMatrix` to it.

widgets/heatmap/heatmap.go
fgksgf: I renamed `v` to `vals`, but don't know how to deal with `xl` and `yl`. `xLabels` and `yLabels` have been used. Do you have any idea ?

mum4k:
Closing for inactivity for now. Please feel free to reopen when you get back to this.

Issue #293, opened 2021-02-12
Circular / Ring Buffer for Text Widget


timbutler:
I've started using this for a quick internal project, with one of features being log files monitored and displayed in real-time while other stats are displayed.

While this is working, one issue we've of course hit is when verbose logs are monitored that the size of the text buffer keeps increasing and of course leads to high memory usage and performance degradation.  

A possible fix to this is to implement a simple ring buffer, where the Text `content` has an (optional) maximum it can grow to. During the `Write` call it could then pop content off the top of the buffer as required so that it never grows further than the maximum content.

I've added this as a basic proof of concept internally, but wanted to confirm if this would be the best approach (or if I'm missing something obvious!) before submitting any pull request.

mum4k:
Hi @timbutler, you raise a good point about the memory usage when the `content` of the `Text` widget just grows.

Adding an optional circular buffer sounds like a great idea. Such PR would be a great addition to the `Text` widget, if you're willing to submit it. Please see [CONTRIBUTING.md](https://github.com/mum4k/termdash/blob/master/CONTRIBUTING.md) for additional instructions (e.g. which branch to fork from).

Don't hesitate to let me know if you have any questions or would like to discuss the API change upfront.

timbutler:
Thanks for the rapid feedback @mum4k! I'll tidy my quick edits and submit a PR in the next few days for further feedback.

mum4k:
Sounds great, looking forward to it @timbutler.

PR #294, opened 2021-02-15
Implements a buffer limit for the Text widget

timbutler:
See issue #293 where memory and performance can degrade with a high number of lines written to the Text widget. 

This is a very simplistic implementation to limit the possible length the text buffer can grow to with the `maxContent` option. 

Default value of -1 means there's no limit and therefore behaviour should remain standard.

It has been working in our test app and allows the use of the Text widget to monitor logs (ie tail) and therefore doesn't bloat over time, but happy to adjust as required.

### DIFF
--- a/widgets/text/options.go
+++ b/widgets/text/options.go
@@ -178,12 +178,14 @@ func ScrollKeys(up, down, pageUp, pageDown keyboard.Key) Option {
 }
 
 // The default for the maximum buffer length within a content area
-// -1 sets as no limit
+// -1 sets as no limit, for logs you may wish to try 10,000 or higher
 const (
 	DefaultMaxContent = -1
 )
 
-// MaxContent - Limits the maximum content within text widget buffer
+// MaxContent - Limits the maximum content within text widget buffer.
+// This is useful when sending large amounts of text to the Text widget, eg
+// when tailing logs as it will limit the memory usage.
 func MaxContent(max int) Option {
 	return option(func(opts *options) {
 		opts.maxContent = max
--- a/widgets/text/options.go
+++ b/widgets/text/options.go
@@ -36,6 +36,7 @@ type options struct {
 	scrollDown       rune
 	wrapMode         wrap.Mode
 	rollContent      bool
+	maxContent       int
 	disableScrolling bool
 	mouseUpButton    mouse.Button
 	mouseDownButton  mouse.Button
@@ -56,6 +57,7 @@ func newOptions(opts ...Option) *options {
 		keyDown:         DefaultScrollKeyDown,
 		keyPgUp:         DefaultScrollKeyPageUp,
 		keyPgDown:       DefaultScrollKeyPageDown,
+		maxContent:      DefaultMaxContent,
 	}
 	for _, o := range opts {
 		o.set(opt)
@@ -172,5 +174,18 @@ func ScrollKeys(up, down, pageUp, pageDown keyboard.Key) Option {
 		opts.keyDown = down
 		opts.keyPgUp = pageUp
 		opts.keyPgDown = pageDown
+	})
+}
+
+// The default for the maximum buffer length within a content area
+// -1 sets as no limit
+const (
+	DefaultMaxContent = -1
+)
+
+// MaxContent - Limits the maximum content within text widget buffer
+func MaxContent(max int) Option {
+	return option(func(opts *options) {
+		opts.maxContent = max
 	})
 }
--- a/widgets/text/text.go
+++ b/widgets/text/text.go
@@ -108,18 +108,24 @@ func (t *Text) Write(text string, wOpts ...WriteOption) error {
 	if opts.replace {
 		t.reset()
 	}
-	for _, r := range text {
-		// Implement basic buffer limit to avoid excess memory issues
-		if len(t.content) > 1 && len(t.content) >= t.opts.maxContent && t.opts.maxContent > 0 {
-			// If the new content is longer than all existing, simply reset
-			if len(string(r)) >= len(t.content) {
-				t.reset()
-			} else {
-				// Truncate by the space required for the new entry
-				t.content = t.content[1:]
-			}
-		}
 
+	// If set, limit the buffer (if maxContent has been set)
+	if len(t.content) >= t.opts.maxContent && t.opts.maxContent > 0 {
+		// If the new content is longer than all existing, simply reset
+		if len(text) >= len(t.content) {
+			t.reset()
+		} else {
+			// Truncate by the space required for the new entry
+			t.content = t.content[len(text):]
+		}
+	}
+
+	// Truncate text as well if it's greater than the maxContent size (if maxContent has been set)
+	if len(text) > t.opts.maxContent && t.opts.maxContent > 0 {
+		text = text[len(text)-t.opts.maxContent:]
+	}
+
+	for _, r := range text {
 		t.content = append(t.content, buffer.NewCell(r, opts.cellOpts))
 	}
 	t.contentChanged = true
--- a/widgets/text/text.go
+++ b/widgets/text/text.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/mum4k/termdash/private/canvas"
 	"github.com/mum4k/termdash/private/canvas/buffer"
+	"github.com/mum4k/termdash/private/runewidth"
 	"github.com/mum4k/termdash/private/wrap"
 	"github.com/mum4k/termdash/terminal/terminalapi"
 	"github.com/mum4k/termdash/widgetapi"
@@ -111,18 +112,20 @@ func (t *Text) Write(text string, wOpts ...WriteOption) error {
 
 	// If set, limit the buffer (if maxContent has been set)
 	if len(t.content) >= t.opts.maxContent && t.opts.maxContent > 0 {
+		fmt.Printf("len: %d strlen: %d", len(t.content), len(t.content))
 		// If the new content is longer than all existing, simply reset
 		if len(text) >= len(t.content) {
 			t.reset()
 		} else {
 			// Truncate by the space required for the new entry
-			t.content = t.content[len(text):]
+			t.content = t.content[runewidth.StringWidth(text):]
 		}
 	}
 
 	// Truncate text as well if it's greater than the maxContent size (if maxContent has been set)
-	if len(text) > t.opts.maxContent && t.opts.maxContent > 0 {
-		text = text[len(text)-t.opts.maxContent:]
+	if runewidth.StringWidth(text) > t.opts.maxContent && t.opts.maxContent > 0 {
+		textrunes := []rune(text)
+		text = string(textrunes[len(textrunes)-t.opts.maxContent:])
 	}
 
 	for _, r := range text {
--- a/widgets/text/text.go
+++ b/widgets/text/text.go
@@ -109,6 +109,17 @@ func (t *Text) Write(text string, wOpts ...WriteOption) error {
 		t.reset()
 	}
 	for _, r := range text {
+		// Implement basic buffer limit to avoid excess memory issues
+		if len(t.content) > 1 && len(t.content) >= t.opts.maxContent && t.opts.maxContent > 0 {
+			// If the new content is longer than all existing, simply reset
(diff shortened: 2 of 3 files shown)


widgets/text/options.go
```diff
@@ -174,3 +176,16 @@ func ScrollKeys(up, down, pageUp, pageDown keyboard.Key) Option {
...
 	})
 }
+
+// The default for the maximum buffer length within a content area
+// -1 sets as no limit
+const (
+	DefaultMaxContent = -1
+)
+
+// MaxContent - Limits the maximum content within text widget buffer
```
mum4k, line 186: This is the only public place where this feature is documented. Would you mind expanding the comment and explaining the behavior `Text` has when this is set?

It would also be good to specify whether we are limiting the number of runes, or the number of cells. I assume this is limiting the number of runes to be displayed and therefore the number of cells may be variable (since some unicode runes take multiple cells).

widgets/text/text.go
```diff
@@ -109,6 +109,17 @@ func (t *Text) Write(text string, wOpts ...WriteOption) error {
 		t.reset()
 	}
 	for _, r := range text {
+		// Implement basic buffer limit to avoid excess memory issues
+		if len(t.content) > 1 && len(t.content) >= t.opts.maxContent && t.opts.maxContent > 0 {
```
mum4k, line 113: Can you help me understand the first condition `len(t.content) > 1`?

Why do we need it, or in other words wouldn't the rest of the condition be enough? I.e.

if len(t.content) >= t.opts.maxContent && t.opts.maxContent > 0 { ... }

widgets/text/text.go
```diff
@@ -109,6 +109,17 @@ func (t *Text) Write(text string, wOpts ...WriteOption) error {
...
+			// If the new content is longer than all existing, simply reset
+			if len(string(r)) >= len(t.content) {
```
mum4k, line 115: I may be missing something, but since `r` is a single rune in this for loop, doesn't the expression `len(string(r))` always return 1?

widgets/text/text.go
```diff
@@ -109,6 +109,17 @@ func (t *Text) Write(text string, wOpts ...WriteOption) error {
 		t.reset()
 	}
 	for _, r := range text {
+		// Implement basic buffer limit to avoid excess memory issues
+		if len(t.content) > 1 && len(t.content) >= t.opts.maxContent && t.opts.maxContent > 0 {
+			// If the new content is longer than all existing, simply reset
```
mum4k, line 114: How do we handle the case when a single `text` given to `Write` is longer than the configured `MaxContent`?

widgets/text/text_test.go
```diff
@@ -809,6 +809,66 @@ func TestTextDraws(t *testing.T) {
...
+			},
+		},
+		{
+			desc:   "tests maxContent length not being limited",
+			canvas: image.Rect(0, 0, 72, 1),
+			opts: []Option{
+				RollContent(),
+			},
+			writes: func(widget *Text) error {
+				return widget.Write("1234567890abcdefghijklmnopqrstuvwxyz1234567890abcdefghijklmnopqrstuvwxyz")
```
mum4k, line 839: Thank you for including test coverage.

The selected test data (content) is fairly long, which makes it harder to reason about what they are testing. Can we minimize its length to the base cases? (a content of 1, max content set to 1, etc...). Using shorter content and MaxContent should arguably be the same behavior as with longer values, but the tests should be easier to follow and verify.

Once we do that, let's make sure we cover all the cases:
- adding content and fitting within MaxContent exactly
- adding content and overwriting portion of the buffer
- adding content and overwriting the full buffer
- adding content that is itself longer than MaxContent

This is the full list that comes to mind. On top of this we should keep the test with newlines and also add at least one test that sets MaxContent and uses unicode characters that are wider than a single cell.

Please let me know if I can provide additional clarification.

mum4k (changes requested):
Thank you for the contribution @timbutler, leaving a few questions.

widgets/text/options.go
timbutler: I've increased the commenting but it may require further clarification here. If my read of the code is correct, `t.content` is the cells so the limits should be based on the cells?

widgets/text/text.go
timbutler: I've fixed this, you are correct in that it isn't required.

widgets/text/text.go
timbutler: I've moved the logic to be outside of the stepping through the runes within the string so it should make more sense again.

widgets/text/text.go
timbutler: I've now fixed this case.

widgets/text/text_test.go
timbutler: I'd simply gone for a longer string for testing just so it stood out and had some variance. As you've pointed out through, it's probably not helpful anyway.

I'll increase the test coverage and commit once ready, I agree that there's extra scenarios to test.

Thanks again for the feedback!

mum4k: Happy to provide the feedback, the thanks goes to you for the contribution!

timbutler: Unicode chars did force a bit of a rethink of the logic to perfect, but the code commit a few days ago should now be correct and the tests (hopefully) enough to cover all cases.

The CI test failure doesn't appear to be related to my changes, please let me know if there's anything else required here.

widgets/text/options.go
mum4k: That is correct, `t.content` is indeed in cells. How would you feel about renaming the option and its default to make it obvious? Something like:

`MaxContentCells` and `DefaultMaxContentCells`.

We could also mention that the number represents the maximum amount of terminal cells the content can take and note that some runes might take multiple cells for clarity.

widgets/text/text.go
```diff
@@ -108,6 +109,25 @@ func (t *Text) Write(text string, wOpts ...WriteOption) error {
 	if opts.replace {
 		t.reset()
 	}
+
+	// If set, limit the buffer (if maxContent has been set)
+	if len(t.content) >= t.opts.maxContent && t.opts.maxContent > 0 {
+		fmt.Printf("len: %d strlen: %d", len(t.content), len(t.content))
```
mum4k, line 115: Was this a debug print that we want to remove before we merge this in?

widgets/text/text.go
```diff
@@ -108,6 +109,25 @@ func (t *Text) Write(text string, wOpts ...WriteOption) error {
 	if opts.replace {
 		t.reset()
 	}
+
+	// If set, limit the buffer (if maxContent has been set)
```
mum4k, line 113: This latest version looks great, thank you for iterating on it. I wonder if we could make the Write function more readable by extracting all code related to content truncation into its own method. Something like:

```go
// Truncates the content and the new text so that the configured MaxContentCells
// is enforced after a Write. 
func (t *Text) truncateToMaxCells(text string) { ... }
```

The `Write` method would then just conditionally call the `truncateToMaxCells` method if `opts.maxContentCells` was set. This might make the "business logic" of `Write` easier to follow.

```go
if t.opts.maxContent > 0 {
  t.truncateToMaxCells(text)
}
```

widgets/text/text.go
```diff
@@ -108,6 +109,25 @@ func (t *Text) Write(text string, wOpts ...WriteOption) error {
...
+	if len(t.content) >= t.opts.maxContent && t.opts.maxContent > 0 {
+		fmt.Printf("len: %d strlen: %d", len(t.content), len(t.content))
+		// If the new content is longer than all existing, simply reset
+		if len(text) >= len(t.content) {
+			t.reset()
+		} else {
+			// Truncate by the space required for the new entry
+			t.content = t.content[runewidth.StringWidth(text):]
```
mum4k, line 121: The calls to `runewidth.StringWidth` do take some work (at the very least O(n) scan over the text. Any way we could refactor this so that the code calls `runewidth.StringWidth(text)` only once?

widgets/text/text.go
```diff
@@ -108,6 +109,25 @@ func (t *Text) Write(text string, wOpts ...WriteOption) error {
 	if opts.replace {
 		t.reset()
 	}
+
+	// If set, limit the buffer (if maxContent has been set)
+	if len(t.content) >= t.opts.maxContent && t.opts.maxContent > 0 {
```
mum4k, line 114: What do we do when someone sets the option to zero cells? Should that be an error when the option is being set?

Can we also add a test case?

widgets/text/text_test.go
```diff
@@ -809,6 +809,112 @@ func TestTextDraws(t *testing.T) {
...
+					c,
+					"12345",
+					image.Point{0, 0},
+				)
+				testcanvas.MustApply(c, ft)
+				return ft
+			},
+		},
+		{
+			desc:   "tests maxContent partial bufffer replacement",
```
mum4k, line 856: Typo: "bufffer".

widgets/text/text_test.go
```diff
@@ -809,6 +809,112 @@ func TestTextDraws(t *testing.T) {
...
+				c := testcanvas.MustNew(ft.Area())
+				// \n still counts as a chacter in the string length
+				testdraw.MustText(c, "ine3", image.Point{0, 0})
+				testdraw.MustText(c, "line4", image.Point{0, 1})
+				testcanvas.MustApply(c, ft)
+				return ft
+			},
+		},
+		{
+			desc:   "tests maxContent exact length of 5",
```
mum4k, line 833: Other interesting test cases are:

- length is just one under the set max
- length is just one over the set max
- multiple writes, the last write causes the content to be reset because the provided text is longer than the configured max
- multiple writes, the last writes truncates the existing content before writing
- multiple writes, the last write truncates text but leaves content untouched

We can also take a look at the coverage report to see which lines aren't covered.

mum4k, line 833: We should also consider what happens with unicode characters that are multi-cell and add a test case where the max content "cuts a rune in half" to verify that Termdash doesn't attempt to print half of a wide rune, but correctly skips it.

Unicode never fails to provide additional entertainment.

mum4k (changes requested):
Looks great, left a few more comments and looks like we could still add more test cases.

Please feel free to let me know if you would prefer to merge this in and have me implement changes based on any of the listed comments. I am happy to help.

Fyi, the CI failure was indeed unrelated to this PR and should now be resolved (after you merge the devel branch).

widgets/text/text.go
timbutler: Yes, just debug which crept in.

widgets/text/text.go
timbutler: Happy for this to be moved out. Happy if you wanted to make these changes, the incoming `text` also needs to be truncated if required as well.

widgets/text/text.go
timbutler: Unless my logic is bad, this should only be called twice during the whole Write procedure. I haven't profiled how intensive it is so if it's worth calling once then happy for this to be refactored.

widgets/text/text.go
timbutler: Easiest fix is probably to set the conditional logic to >= 0. It'll then be ignored since the only valid size is 1 or greater.

widgets/text/text_test.go
timbutler: I believe this is covered in the test `tests maxContent partial bufffer replacement`, as it has to truncate based on a character not the data (hence starting with 你 for the result). I was using this as a test case for this as I've never worked with Go for unicode before!

If I get the time I'll make some of the adjustments later in the week, otherwise happy if you want to merge and correct as well.

mum4k:
Thank you very much @timbutler, I will be happy to merge as is and work on the remaining items.

Before we do that I have one more CLA related favor to ask for. Can you please explicitly state that you give me permission to merge this code into the master branch? This is required to pass the CLA check when I will be releasing the commits authored by you.

timbutler:
Hi @mum4k, sorry for the delay. I'm happy for all of my code to be merged into the master branch.

mum4k:
Thank you again for the confirmation and the contribution @timbutler.

mum4k (approved):

PR #342, opened 2022-12-08
Gauge threshold

lovromazgon:
This PR introduces the option `gauge.Threshold` which allows you to draw a vertical line on a gauge representing a threshold. If the gauge is drawn with a border the threshold is drawn over the border, otherwise it's only drawn within the gauge itself. The threshold is only drawn if its value is within the gauge (greater than 0 and less than `g.total`).

I took the liberty to change the demo to draw a threshold on two of the gauges.
image


mum4k:
This looks great, thank you @lovromazgon. I will review and respond in the next few days.

mum4k:
Close/reopen to trigger CI.

widgets/gauge/options.go
```diff
@@ -201,3 +205,17 @@ func BorderTitleAlign(h align.Horizontal) Option {
 		opts.borderTitleHAlign = h
 	})
 }
+
+// Threshold configures the Gauge to display a vertical threshold line at value
+// t. If the progress is set by a call to Percent(), t represents a percentage,
+// e.g. "40" means line is displayed at 40%. If the progress is set by a call to
```
mum4k, line 211: One of the principles I was trying to follow in Termdash is that invalid configuration will notify the user rather than silently take and ignore the value. 

How would you feel about modifying this, so that it returns an error if the user sets the threshold to less than 0? We could reuse the existing `validate` function.

widgets/gauge/gauge.go
```diff
@@ -284,6 +286,25 @@ func (g *Gauge) Draw(cvs *canvas.Canvas, meta *widgetapi.Meta) error {
 			return err
 		}
 	}
+	if g.opts.threshold > 0 && g.opts.threshold < g.total {
```
mum4k, line 289: In order to keep the Draw() function focused on higher level logic - could we move this new code into private helper methods? So that the resulting code here would look something like this:

```go
  if g.thresholdVisible() {
    if err := g.drawThreshold(...); err != nil { ... }
  }
```

widgets/gauge/gauge_test.go
```diff
@@ -841,6 +841,100 @@ func TestGauge(t *testing.T) {
...
+				testdraw.MustRectangle(c, image.Rect(0, 0, 3, 3),
+					draw.RectChar('o'),
+					draw.RectCellOpts(cell.BgColor(cell.ColorGreen)),
+				)
+				testcanvas.MustApply(c, ft)
+				return ft
+			},
+		},
+		{
+			desc: "threshold outside of bounds (>=max)",
```
mum4k, line 918: Can we also add a test cases based on where the progress is relative to the threshold. E.g.:

- progress below threshold
- progress exactly at threshold
- progress after threshold

mum4k (changes requested):
The code looks great, thank you for including good documentation and tests. I only have a few questions / suggestions.

widgets/gauge/gauge_test.go
```diff
@@ -70,9 +70,9 @@ func TestGauge(t *testing.T) {
 			wantErr: true,
 		},
 		{
-			desc: "fails on negative height",
+			desc: "fails on negative threshold",
```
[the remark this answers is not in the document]
mum4k: Sounds good, thank you.

mum4k (approved):
Thank you @lovromazgon!

PR #347, opened 2023-01-29
Add OnChange handler to the TextInput widget

LQR471814:
This PR adds an `onChange` option to the TextInput widget. This is an event handler similar to the existing `OnSubmit`, it is called whenever a character is inserted or deleted from the TextInput's text content.

Rudimentary tests have been written and an example of it's usage is given in `textinputdemo.go`.

### DIFF
--- a/widgets/textinput/editor_test.go
+++ b/widgets/textinput/editor_test.go
@@ -330,6 +330,7 @@ func TestFieldEditor(t *testing.T) {
 		wantContent string
 		wantCurIdx  int
 		wantErr     bool
+		mutations   int
 	}{
 		{
 			desc:    "fails for width too small",
@@ -355,6 +356,7 @@ func TestFieldEditor(t *testing.T) {
 			wantView:    "abc",
 			wantContent: "abc",
 			wantCurIdx:  3,
+			mutations:   3,
 		},
 		{
 			desc:  "longer data than the width, cursor at the end",
@@ -369,6 +371,7 @@ func TestFieldEditor(t *testing.T) {
 			wantView:    "⇦cd",
 			wantContent: "abcd",
 			wantCurIdx:  3,
+			mutations:   4,
 		},
 		{
 			desc:  "longer data than the width, cursor at the end, has full-width runes",
@@ -383,6 +386,7 @@ func TestFieldEditor(t *testing.T) {
 			wantView:    "⇦世",
 			wantContent: "abc世",
 			wantCurIdx:  3,
+			mutations:   4,
 		},
 		{
 			desc:  "width decreased, adjusts cursor and shifts data",
@@ -400,6 +404,7 @@ func TestFieldEditor(t *testing.T) {
 			wantView:    "⇦cd",
 			wantContent: "abcd",
 			wantCurIdx:  3,
+			mutations:   4,
 		},
 		{
 			desc:  "cursor won't go right beyond the end of the data",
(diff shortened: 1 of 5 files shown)


widgets/textinput/editor_test.go
```diff
@@ -330,6 +330,7 @@ func TestFieldEditor(t *testing.T) {
 		wantContent string
 		wantCurIdx  int
 		wantErr     bool
+		mutations   int
```
mum4k, line 333: The name mutations does not make me think of the onChange handler. Can we choose a more closely related name. E.g. we could call it wantOnChangeCalls instead.

widgets/textinput/editor_test.go
```diff
@@ -1808,6 +1881,10 @@ func TestFieldEditor(t *testing.T) {
 			if gotContent != tc.wantContent {
 				t.Errorf("content -> %q, want %q", gotContent, tc.wantContent)
 			}
+
+			if tc.mutations != changeCount {
+				t.Errorf("mutation count -> %d, want %d", changeCount, tc.mutations)
```
mum4k, line 1886: Once we call it wantOnChangeCalls, we can also adjust this error message.

```go
t.Errorf("unexpected number of onChange calls -> %d, want %d", ...)
```

widgets/textinput/options.go
```diff
@@ -269,6 +270,19 @@ func OnSubmit(fn SubmitFn) Option {
 	})
 }
 
+// ChangeFn if provided is called when the content of the text input field changes,
+// the argument data contains all the text in the field.
```
mum4k, line 274: Let's clarify that the argument contains the content **after** the change.

widgets/textinput/options.go
```diff
@@ -269,6 +270,19 @@ func OnSubmit(fn SubmitFn) Option {
...
+//
+// The callback function must be thread-safe as the keyboard event that
+// triggers the submission comes from a separate goroutine.
+type ChangeFn func(data string)
+
+func OnChange(fn ChangeFn) Option {
```
mum4k, line 280: Let's document this option as well, since it is public and exported.

We could probably split the above comment into two. Leave the portion that describes the argument of the ChangeFn next to it and describe the option itself here.

widgets/textinput/textinput.go
```diff
@@ -73,7 +73,7 @@ func New(opts ...Option) (*TextInput, error) {
 		editor: newFieldEditor(),
 		opts:   opt,
 	}
-
+	ti.editor.onChange = opt.onChange
```
mum4k, line 76: This introduces a slight layering problem, since the constructor of the TextInput object now understand details of how to construct another object (the field editor).

Can we modify the newFieldEditor() function instead, giving an argument of type OnChangeFn and pass it down on line 73 where newFieldEditor() is called?

mum4k (changes requested):
Thank you for the contribution @LQR471814!

Please take a look at the tests, since some of them seem to be failing. Feel free to ignore the coverage error at the end, that is something that I need to fix.

mum4k:
One more request, can you please give me an explicit permission to merge this code into the master branch when releasing the next version? I will need that to bypass the CLA check at that point.

LQR471814:
I give you explicit permission to merge this code into the master branch and do whatever you want with it.

mum4k:
I apologize for the delay, I did not notice the comments were already addressed. I will look at this in the next few days.

mum4k (approved):
Looks good, thank you for the contribution @LQR471814 !

Issue #358, opened 2023-12-28
How to add sub-container with id 


pavel-blagodov:
Hey,

Could you please clarify how to add sub-container with id, so that it can be updated without updating siblings containers. I'm using grid layout

Thank you

mum4k:
Thank you for reaching out @pavel-blagodov. To update a sub-container, you would first give that sub-container a known ID using the `contained.ID("some_id")` option. You can then update any options on that container including sub-splits by calling the `container.Update()` method as documented in [Dynamic Layout](https://github.com/mum4k/termdash/wiki/Dynamic-layout) and referring back to the same `ID`.

Please refer to https://github.com/mum4k/termdash/pull/359 which demonstrates this on the `termdashdemo`. When running the demo, you can press `Ctrl-U` to update a sub-container (changes widget from green sparkline to a red sparkline). Any other updates are possible as well.

PR #359, opened 2024-01-04
Demonstrate updating a sub-container.

mum4k:


mum4k:
Example to be used in #358

Issue #360, opened 2024-01-11
Panic when attempting to use Backtab


spacez320:
Pressing the 'backtab' key (a.k.a. Shift + Tab) throws an error in Termdash.

```
panic: unknown keyboard key '278' in a keyboard event Backtab
goroutine 31 [running]:
  internal/lib.errorTermdashHandler({0x95a4c0?, 0xc0016f5af0})
    /home/matthew/workspace/personal/cryptarch/internal/lib/display_termdash.go:59 +0x75
  github.com/mum4k/termdash.(*termdash).handleError(...)
    /home/matthew/lib/go/pkg/mod/github.com/mum4k/termdash@v0.18.0/termdash.go:250
  github.com/mum4k/termdash.(*termdash).subscribers.func1({0x95a660?, 0xc0016f5ad0?})
    /home/matthew/lib/go/pkg/mod/github.com/mum4k/termdash@v0.18.0/termdash.go:215 +0xbc
  github.com/mum4k/termdash/private/event.(*subscriber).callback(0xc00033e200, {0x95a660?, 0xc0016f5ad0?})
    /home/matthew/lib/go/pkg/mod/github.com/mum4k/termdash@v0.18.0/private/event/event.go:95 +0x31
  github.com/mum4k/termdash/private/event.(*subscriber).run(0xc00033e200, {0x95d8f0, 0xc0000940a0})
    /home/matthew/lib/go/pkg/mod/github.com/mum4k/termdash@v0.18.0/private/event/event.go:110 +0x5a
  created by github.com/mum4k/termdash/private/event.newSubscriber
    /home/matthew/lib/go/pkg/mod/github.com/mum4k/termdash@v0.18.0/private/event/event.go:89 +0x28d
```

This is a supported key in Tcell, at least. https://godocs.io/github.com/gdamore/tcell/v2#Key

Currently I have my program throwing away the error with a handler, but would be nice to properly support the key press.

spacez320:
@mum4k Thanks for the response. I'd be happy to contribute this. Let me know if there's a time table, but otherwise should be able to report back soon.

mum4k:
Thank you @spacez320, no rush - I don't keep any regular release schedule for Termdash. We can tag a new version once we have your PR in.

Thank you for your help!

spacez320:
Thanks @mum4k . I have a PR here: https://github.com/mum4k/termdash/pull/362, let me know how it looks.

PR #362, opened 2024-01-27
Implement backtab for tcell widgets

spacez320:
This implements "Backtab" (a.k.a. Shift+Tab) as a supported keystroke for tcell widgets. Uses backtab in a few tests and demos as makes sense.


spacez320:
Closes https://github.com/mum4k/termdash/issues/360

mum4k (changes requested):
Looks great, thank you @spacez320.

Please remove the binary file `widgets/textinput/formdemo/formdemo` from the PR. Meanwhile I will verify that all the tests pass and we should be good to go.

spacez320:
@mum4k Removed, sorry about that, and thanks for looking.

mum4k (approved):
Thank you for the contribution!

PR #373, opened 2024-03-05
WIP Allow reverse horizontal area splits

spacez320:


private/area/area.go
```diff
@@ -106,7 +106,7 @@ func VSplitCells(area image.Rectangle, cells int) (left image.Rectangle, right i
 // be a zero or a positive integer. Providing a zero returns top=image.ZR,
 // bottom=area. Providing a number equal or larger to area's height returns
 // top=area, bottom=image.ZR.
-func HSplitCells(area image.Rectangle, cells int) (top image.Rectangle, bottom image.Rectangle, err error) {
+func HSplitCells(area image.Rectangle, cells int, fromMax bool) (top image.Rectangle, bottom image.Rectangle, err error) {
```
mum4k, line 109: While we technically can break these APIs marked as "private", I would recommend we only do that if we need to. This might not be the case.

Additionally adding a boolean flag argument to a function call tends to hurt readability of the call sites. The call sites will look like this:

```Go
HSplitCells(area, cells, true)  // True what?
```

Which naturally brings up the question - "what does `true` mean? Martin Fowler shared some thoughts on boolean flags here: https://martinfowler.com/bliki/FlagArgument.html

My recommendation would be to leave `HSplitCells` as is and add a new `HSplitCellsReversed` function. If `HSplitCells` and `HSplitCellsReversed` end up having a lot of logic in common, we could extract that into a new private function. We could use a similar approach for vertical splits.

Let me know what you think.

mum4k (changes requested):

private/area/area.go
spacez320: No problem @mum4k , makes sense, will defer to this and also keep the API under `private` consistent.

mum4k (approved):
Thank you again for the contribution @spacez320. Appreciate the included test coverage and the fixess to existing docs.

PR #377, opened 2024-06-28
[FEAT] reset for linechart

woodliu:
I create a linechart for prometheus metrics,  and when i change the PromQL in inputtest, i need draw a total different linechart.
For example:

1. First i create a linechart with PromQL: `node_receive_packet_total{}`
2. Seconde i change the PromQL to : `rate(node_receive_packet_total{}[1m])`
That is the total different graph , so i need to reset the current graph ,and draw new one

### DIFF
--- a/widgets/linechart/linechart.go
+++ b/widgets/linechart/linechart.go
@@ -232,6 +232,13 @@ func (lc *LineChart) Series(label string, values []float64, opts ...SeriesOption
 	return nil
 }
 
+func (lc *LineChart) Reset() {
+	if lc == nil {
+		return
+	}
+	lc.series = make(map[string]*seriesValues)
+}
+
 // xDetails returns the details for the X axis given the specified minimum and
 // maximum value to display.
 func (lc *LineChart) xDetails(cvs *canvas.Canvas, reqYWidth, min, max int) (*axes.XDetails, error) {


widgets/linechart/linechart.go
```diff
@@ -232,6 +232,13 @@ func (lc *LineChart) Series(label string, values []float64, opts ...SeriesOption
 	return nil
 }
 
+func (lc *LineChart) Reset() {
```
mum4k, line 235: Thank you for the contribution. We can certainly add a `Reset` method on the `LineChart` widget, but there might be a more elegant solution. If we add it here, we will have to solve this problem on each widget. An alternative is to use the dynamic layout feature in the `container` layer.

With dynamic layout, we can tag a `container` with an arbitrary `ID` and later use that `ID` to reference the `container` and replace its `widget`:

https://github.com/mum4k/termdash/wiki/Dynamic-layout#replacing-widget-in-a-container

So to achieve the desired functionality you can:
- set an `ID` on the target `container` when creating it.
- create a new and empty instance of the `LineChart` widget.
- use the container's `Update` method together with the `ID` and the `container.PlaceWidget` option as in the example linked above.

If you prefer to continue adding the `Reset` method here, we need to do a bit more work:
- make the `Reset` method thread safe by acquiring the mutex on the `LineChart` object.
- reset all the fields, not just `series` (might be simpler to just replace self with a new object).
- document the method.
- add test coverage for the method.

Please let me know how you would like to proceed.

woodliu: Hi, @mum4k i think that would workable, but if i have some widgets in a container, like this:
image
If i just clean only the LineChart data, do i need to do this again to output the new gridOpts?
```
builder.Add(elementFromGraphAndLegend(gr.widgetExprInput, gr.widgetExprLegend, gr.widgetGraph, gr.widgetGraphLegend))
gridOpts, err := builder.Build()
```

mum4k: The same approach works with the Grid layout. The idea is:

1. Mark the container where we want to replace the widget with an ID we know.
2. Update the container when required and use the `container.PlaceWidget` option to replace the old widget with any new widget.

Here is [an example](https://github.com/mum4k/termdash/pull/379/files) of how to achieve `Linechart` widget replacement with the grid layout. Feel free to run the demo and press `r` to replace the main linechart with another one.

mum4k: Going to close this PR, since we have a solution that works. However feel free to reopen and ask for additional details if need be.

Issue #383, opened 2025-08-23
Pie chart or donut multi values


VincenzoManto:
Might be useful to have also a Pie chart or a donut multi values like this:

<img width="1473" height="623" alt="Image" src="https://github.com/user-attachments/assets/6b0f85dc-f410-471d-afd2-0219d2579a3d" />

mum4k:
Really cool, thank you for working on this @VincenzoManto !

PR #384, opened 2025-08-23
Implement Pie widget

VincenzoManto:
As in #383 this is an implementation of a Pie widget (better a donut multi values)

### DIFF
--- a/widgets/pie/options.go
+++ b/widgets/pie/options.go
@@ -26,7 +26,13 @@ type options struct {
 // validates the provided options
 // at the moment no validation is performed cause options are not required
 func (o *options) validate() error {
-	return nil
+	return nil	
+}
+// ColorOption sets custom colors for the pie chart segments.
+func ColorOption(colors []cell.Color) Option {
+	return option(func(opts *options) {
+		opts.colors = colors
+	})
 }
 
 // newOptions creates a new options instance.
--- a/widgets/pie/options.go
+++ b/widgets/pie/options.go
@@ -1,0 +1,50 @@
+package pie
+
+import (
+	"github.com/mum4k/termdash/cell"
+)
+
+// Option defines a function that sets a specific option for the Pie widget.
+type Option interface {
+	// set sets the provided option.
+	set(*options)
+}
+
+// option implements Option.
+type option func(*options)
+
+// set implements Option.set.
+func (o option) set(opts *options) {
+	o(opts)
+}
+
+// options stores the provided options.
+type options struct {
+	colors []cell.Color
+}
+
+// validates the provided options
+// at the moment no validation is performed cause options are not required
+func (o *options) validate() error {
+	return nil
+}
+
+// newOptions creates a new options instance.
+func newOptions() *options {
+	return &options{
+		colors: DefaultColors,
+	}
+}
+
+// DefaultColors defines a default set of colors used for rendering pie chart segments.
+// These colors are chosen from the predefined cell.Color constants and include a variety
+// of primary and secondary colors to ensure visual distinction between segments.
+var DefaultColors = []cell.Color{
+	cell.ColorRed,
+	cell.ColorGreen,
+	cell.ColorBlue,
+	cell.ColorYellow,
+	cell.ColorMagenta,
+	cell.ColorCyan,
+	cell.ColorWhite,
+}
--- a/widgets/pie/options.go
+++ b/widgets/pie/options.go
@@ -26,8 +26,9 @@ type options struct {
 // validates the provided options
 // at the moment no validation is performed cause options are not required
 func (o *options) validate() error {
-	return nil	
+	return nil
 }
+
 // ColorOption sets custom colors for the pie chart segments.
 func ColorOption(colors []cell.Color) Option {
 	return option(func(opts *options) {
--- a/widgets/pie/pie.go
+++ b/widgets/pie/pie.go
@@ -85,7 +85,6 @@ func pieChartMidAndRadii(ar image.Rectangle) (image.Point, int) {
 // the outer radius. The method ensures thread safety by locking the Pie's
 // mutex during the drawing process.
 //
-//
 // The number of colors in the list is not significant. If there are more values than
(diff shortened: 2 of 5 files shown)


widgets/pie/options.go
```diff
@@ -0,0 +1,50 @@
+package pie
```
mum4k, line 1: Would you mind editing the README.md as well? This is where we list all the available widgets, we could list this one too.

https://github.com/mum4k/termdash/blob/master/README.md

widgets/pie/options.go
```diff
@@ -0,0 +1,50 @@
...
+// validates the provided options
+// at the moment no validation is performed cause options are not required
+func (o *options) validate() error {
+	return nil
+}
+
+// newOptions creates a new options instance.
+func newOptions() *options {
+	return &options{
+		colors: DefaultColors,
```
mum4k, line 42: Did you want to add an option to set custom colors?

Example of an option:

https://github.com/mum4k/termdash/blob/102df20a882547528c83047688595acd2a82c7ef/widgets/button/options.go#L103-L108

widgets/pie/options.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+}
+
+// DefaultColors defines a default set of colors used for rendering pie chart segments.
+// These colors are chosen from the predefined cell.Color constants and include a variety
+// of primary and secondary colors to ensure visual distinction between segments.
```
mum4k, line 48: Can we document whether the number of colors in the list is significant and how? E.g. what happens if there are more values than colors? Will they get reused in a round-robin fashion? Something else?

widgets/pie/pie.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	for _, opt := range opts {
+		opt.set(p.opts)
+	}
+
+	p.values = values
+	p.total = 0
+	if len(p.colors) == 0 {
+		p.colors = DefaultColors
```
mum4k, line 54: Typically this should be handled by the initialization of options in the options.go file. It might be redundant here, though it doesn't hurt.

widgets/pie/pie.go
```diff
@@ -0,0 +1,167 @@
...
+	innerRadiusX := int(float64(radiusX) * 0.6)
+	innerRadiusY := int(float64(radiusY) * 0.6)
+
+	currentAngle := 0.0
+	for i, value := range p.values {
+		endAngle := currentAngle + float64(value)/float64(p.total)*2*math.Pi
+		color := p.colors[i%len(p.colors)]
+
+		// I draw a series of radial lines from the inner radius to the outer radius.
+		for angle := currentAngle; angle < endAngle; angle += 0.01 {
```
mum4k: (optional) This is potentially expensive. As an optimization we could consider using the draw library which can dra circles or even partial circles starting and ending on specific angles:

https://github.com/mum4k/termdash/blob/102df20a882547528c83047688595acd2a82c7ef/private/draw/braille_circle.go#L110-L114

https://github.com/mum4k/termdash/blob/102df20a882547528c83047688595acd2a82c7ef/private/draw/braille_circle.go#L88-L99

widgets/pie/pie_test.go
```diff
@@ -0,0 +1,135 @@
...
+			values: []int{10, 20},
+			colors: []cell.Color{cell.ColorRed, cell.ColorBlue},
+			canvas: image.Rect(0, 0, 10, 10),
+		},
+		{
+			desc:        "Fails to draw when canvas is too small",
+			values:      []int{10, 20},
+			colors:      []cell.Color{cell.ColorRed, cell.ColorBlue},
+			canvas:      image.Rect(0, 0, 1, 1),
+			wantDrawErr: true,
```
mum4k, line 58: (optional) Consider adding a few test cases that draw on the unit test only canvas.

Example from the donut widget:

https://github.com/mum4k/termdash/blob/102df20a882547528c83047688595acd2a82c7ef/widgets/donut/donut_test.go#L271-L286

This would likely require using the draw.BrailleCircle function suggested above.

mum4k (changes requested):
Thank you for the contribution!

VincenzoManto:
@mum4k thank you so much for your review. I pushed new changes. Please, tell me if there is something else to correct, I'll fix it

widgets/pie/options.go
mum4k: Thank you for adding the documentation, the added documentation links to image `./doc/images/piedemo.gif` which doesn't seem to exist. Did you mean to add it in this PR?

widgets/pie/pie.go
```diff
@@ -110,29 +109,26 @@ func (p *Pie) Draw(cvs *canvas.Canvas, meta *widgetapi.Meta) error {
 		return fmt.Errorf("braille.New => %v", err)
 	}
 
-	mid, radiusX, radiusY := pieChartMidAndRadii(cvs.Area())
+	mid, radiusX := pieChartMidAndRadii(cvs.Area())
 
 	innerRadiusX := int(float64(radiusX) * 0.6)
-	innerRadiusY := int(float64(radiusY) * 0.6)
+	// innerRadiusY := int(float64(radiusY) * 0.6)
```
mum4k: (nit) If this isn't needed, can we remove the line?

mum4k (approved):
Thank you for making the changes, looks good. We have two remaining comments, both optional) if you want to take a look before we merge this in.

widgets/pie/pie.go
VincenzoManto: Sure

widgets/pie/options.go
VincenzoManto: I removed the GIF from the README.md file (I see that donut, for example, doesn't have a GIF in its documentation). If needed, I'll try creating one 😊

VincenzoManto:
Hi @mum4k, thanks again for your review. If you need anything else, just let me know!

widgets/pie/options.go
mum4k: That is fine, the gif is optional.

mum4k:
Once again, thank you for the contribution.

PR #396, opened 2026-06-30
fix(modal): measure docked title by display width, not rune count

mahirhir:
Minimized draggable windows size and label their dock by the title's rune count, but the title is rendered by `draw.Text` (with `draw.TextMaxX`), which measures display width. Wide runes such as CJK occupy two terminal columns, so a window with a wide title is docked too narrow and its compact label gets clipped.

`dockedWidth` used `len([]rune(" " + title + " "))` and `compactDockTitle` compared and truncated by `len([]rune(...))`. Every other widget that lays out user text (`toast`, `tab`, `dropdown`, `treeview`, `checkbox`, `radio`, `textinput`) measures it with `runewidth.StringWidth` / `runewidth.RuneWidth`, and `treeview.truncateString` already truncates with an ellipsis by display width. This change makes the dock title consistent with that.

Example: a window titled `データベース管理` (8 runes, 16 columns) was docked at width 10 instead of 18, and the compact label for a longer wide title overflowed past the requested width.

Added table-driven regression tests for `dockedWidth` and `compactDockTitle`. ASCII titles are unaffected because their display width equals the rune count; the new wide-title cases fail before this change and pass after. `go build ./...`, `go vet ./...`, and `go test -race ./...` pass.

Per CONTRIBUTING this targets the `devel` branch.


mahirhir:
Closing this to clean up my open PRs. It's a small change and not really worth taking up your review time. Sorry for the noise.

